PR #2, opened 2015-06-17
improved ahead/behind showing

jschneider:
Hi guys,

I have added the code from the original git fish prompt.  I prefer that output.
Attention: The code is not clean. My fish-foo ist very limited and I just hacked it in.

But maybe something of that kind could be added?

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -304,7 +304,6 @@ function __bobthefish_prompt_git -d 'Display the actual git state'
 
 		switch "$count"
     		case '' # no upstream
-    		  echo "no upstream"
     		case "0	0" # equal to upstream
     			set ahead ""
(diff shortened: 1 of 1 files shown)


bobthecow:
You might prefer that version, but is this theme called `johannesthefish`? ;)

This has been merged, but see cd15f6551c7cef433a2e410d0f3733724fd3ec61 for some changes:

 * reverted your removal of the old style (this is `bobthefish`, after all :P).
 * added a flag for toggling between verbose and non-verbose ahead/behind.
 * extracted the code into a couple of helper methods to make it easier to reason about.
 * changed things to be more in line with the usual fish/omf coding style.
 * removed a bunch of unnecessary local variables.

To get the style you prefer, add `set theme_display_git_ahead_verbose yes` to your `config.fish`.

PR #3, opened 2015-06-26
Fix ahead printing when both ahead and behind

tchajed:
Specifically, the awk parsing bails out early if we're ahead, even though there may be later behind commits, resulting in a status of just "+" when "±" would be correct.

I believe the awk nextfile is an optimization to stop parsing early, so I left it in, but only if we already have both a > 0 and b > 0, in which case the printed status can't change.


bobthecow:
Great, thank you. And yeah, that's exactly why nextfile is there.

PR #11, opened 2015-10-13
Title and time

qatsi:
Hi,
I made some customizations.
- Ruby version is now shorter, only gemset is shown
- Ruby version displayed only for Ruby-enabled folders
- Right prompt shows shorten time and week of the day
- Title has abbreviated current directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,10 @@
 # bobthefish
 
-bobthefish is a Powerline-style, Git-aware fish theme optimized for awesome.
+`bobthefish` is a Powerline-style, Git-aware [fish][fish] theme optimized for awesome.
 
 [![](https://img.shields.io/badge/Framework-Oh My Fish-blue.svg?style=flat)](https://github.com/oh-my-fish/oh-my-fish) ![](https://img.shields.io/cocoapods/l/AFNetworking.svg) [![Join the chat at https://gitter.im/oh-my-fish/oh-my-fish](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/oh-my-fish/oh-my-fish?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
 ![bobthefish][screenshot]
-
 
 ### Installation
 
@@ -18,7 +17,6 @@ You will probably need a [Powerline-patched font][patching] for this to work.
 
 This theme is based loosely on [agnoster][agnoster].
 
-
 ### Features
 
  * A helpful, but not too distracting, greeting.
@@ -28,7 +26,6 @@ This theme is based loosely on [agnoster][agnoster].
  * An abbreviated path which doesn't abbreviate the name of the current project.
  * All the things you need to know about Git in a glance.
  * Visual indication that you can't write to the current directory.
-
 
 ### The Prompt
 
@@ -54,10 +51,9 @@ This theme is based loosely on [agnoster][agnoster].
      * _Note that not all of these have been implemented for hg yet :)_
  * Abbreviated project-relative path
 
-
 ### Configuration
 
-You can override some default options in your `config.fish`:
+You can override some of the following default options in your `config.fish`:
 
 ```fish
 set -g theme_display_git no
@@ -69,12 +65,22 @@ set -g theme_display_ruby no
 set -g theme_display_user yes
 set -g theme_title_display_process yes
 set -g theme_title_display_path no
+set -g theme_title_use_abbreviated_path no
 set -g theme_date_format "+%a %H:%M"
 set -g theme_avoid_ambiguous_glyphs yes
 set -g default_user your_normal_user
 ```
 
+**Title options**
 
+- `theme_title_display_process`. By default theme doesn't show current process name in terminal title. If you want to show it, just set to `yes`.
+- `theme_title_display_path`. Use `no` to hide current working directory from title.
+- `theme_title_use_abbreviated_path`. Default is `yes`. This means your home directory will be displayed as `~` and `/usr/local` as `/u/local`. Set it to `no` if you prefer full paths in title.
+
+**Prompt options**
+- `theme_display_ruby`. Use `no` to completely hide all information about Ruby version. By default Ruby version displayed if there is the difference from default settings.
+
+[fish]:       https://github.com/fish-shell/fish-shell
 [screenshot]: http://i.0x7f.us/bobthefish.png
 [patching]:   https://powerline.readthedocs.org/en/latest/fontpatching.html
 [fonts]:      https://github.com/Lokaltog/powerline-fonts
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -2,7 +2,7 @@
 #
 # bobthefish is a Powerline-style, Git-aware fish theme optimized for awesome.
 #
-# You will probably need a Powerline-patched font for this to work:
+# You will need a Powerline-patched font for this to work:
 #
 #     https://powerline.readthedocs.org/en/latest/fontpatching.html
 #
@@ -10,7 +10,7 @@
 #
 #     https://github.com/Lokaltog/powerline-fonts
 #
-# You can override some default options in your config.fish:
+# You can override some default prompt options in your config.fish:
 #
 #     set -g theme_display_git no
 #     set -g theme_display_git_untracked no
@@ -19,8 +19,6 @@
 #     set -g theme_display_virtualenv no
 #     set -g theme_display_ruby no
 #     set -g theme_display_user yes
-#     set -g theme_title_display_process yes
-#     set -g theme_title_display_path no
 #     set -g theme_date_format "+%a %H:%M"
 #     set -g theme_avoid_ambiguous_glyphs yes
 #     set -g default_user your_normal_user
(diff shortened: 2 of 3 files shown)


bobthecow:
These are great ideas, but they generally just customize the theme to your liking. I'd prefer to make the whole theme a bit more customizable instead :)

How about we add a few theme customization variables:

```
set -g theme_title_display_process no
set -g theme_title_display_path no
set -g theme_date_format "+%a %H:%M"
```

With this change:

 * Users with the process already in the title (like iTerm 2 does) can suppress it, but other users still have a process shown.
 * Users who want no path in the title can also remove it, if that's a thing they're into.
 * Rather than hardcoding one arbitrary date format instead of another, let's expose that as a customization variable to allow users to choose their own format.

Then let's use the fish builtin `prompt_pwd` for the title to keep things simpler, since it's basically the same output as your abbreviated directory code.

That covers everything except the ruby version output customization. I'm not quite sure the best way to make that work, though. The three arguments were there for a reason. Showing just the gemset is problematic for users who actually switch between different interpreter versions, or ruby version numbers, or who don't use named gemsets. It also breaks ruby version support in all sorts of ways, because of your assumption that there will be a `.ruby-version` file in the current directory. What's a good way to let you show just what you want, and let others show the more canonical rvm version strings?

qatsi:
Thanks for advice.
I also changed code for title as proposed. By default it behaves as same as *before this pull request* i.e. shows process name ($_) and full $PWD. So current users will see difference only if they tweak their `config.fish`.

- `theme_title_display_process` **no** to hide
- `theme_title_display_path` **no** to hide
- `theme_title_use_abbreviated_path` **yes** to use abb abbreviated path

Please take a look in `README.md` where I’ve added description for each of setting.
If compare to mentioned 13c35f34248530ec66131a9c4498ce08b0893f49 my solution may be used in any combinations e.g. only process name or only current directory or both without extra space.
If compare to e71317aa1982691ee85eac49cce6bc8222c9db77 my solution let user choose which path format is preferable. Like yours it uses `prompt_pwd` for abbreviated directory name.

qatsi:
Hi, I've just finished customizations. Please take a look.

fish_title.fish
```diff
@@ -1,4 +1,19 @@
...
-  [ "$theme_title_display_process" = 'yes' ]; and echo $_ ' '
-  [ "$theme_title_display_path" != 'no' ]; and prompt_pwd
+  if [ "$theme_title_display_process" != 'no' ]
+    echo $_
+    [ "$theme_title_display_path" != 'no' ]; and echo ' '
+  end
+
+  if [ "$theme_title_display_path" != 'no' ]
+    if [ "$theme_title_use_abbreviated_path" = 'yes' ]
+      echo (prompt_pwd)
```
bobthecow: you don't need `echo (prompt_pwd)` here, just `prompt_pwd` … because it echos by default :)

fish_prompt.fish
```diff
@@ -407,6 +406,17 @@ function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
   set_color normal
 end
 
+function __bobthefish_prompt_rubies -d 'Display current Ruby information'
+  switch "$theme_display_ruby"
+    case no; return
+    case auto
+      if [ -f '.ruby-version' -o -f '.ruby-version' -o -f 'Gemfile' -o \
```
bobthecow: why two `.ruby-version` here?

fish_prompt.fish
```diff
@@ -407,6 +406,17 @@ function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
   set_color normal
 end
 
+function __bobthefish_prompt_rubies -d 'Display current Ruby information'
+  switch "$theme_display_ruby"
+    case no; return
+    case auto
```
bobthecow: What exactly are you trying to accomplish by only showing the ruby version inside the root directory affected by a version switch? If you run `ruby` inside a subdirectory, it'll be affected by the `.ruby-version` as well…

fish_prompt.fish
```diff
@@ -391,11 +389,12 @@ function __bobthefish_prompt_virtualfish -d "Display activated virtual environme
...
 end
 
-function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
-  [ "$theme_display_ruby" = 'no' ]; and return
-  if which rvm-prompt >/dev/null 2>&1
-    set ruby_version (rvm-prompt i v g)
-  else if which rbenv >/dev/null 2>&1
+function __bobthefish_show_ruby -d 'Current Ruby (rvm/rbenv)'
+  set -l ruby_version
+  set -q theme_display_ruby_format; or set -l theme_display_ruby_format i v g
```
bobthecow: since this only applies to rvm, let's make the option `theme_display_rvm_format` and move it into the `if type rvm-prompt`

fish_prompt.fish
```diff
@@ -391,11 +389,12 @@ function __bobthefish_prompt_virtualfish -d "Display activated virtual environme
...
+  if type rvm-prompt >/dev/null 2>&1
```
bobthecow: Changing these back to `type` from `which` is a regression (likely from a merge conflict resolution). Please keep the `which` version :)

bobthecow:
I think not displaying the process and using the abbreviated path in the title are better defaults, so let's switch the `theme_title_display_process` logic back and change `theme_title_use_abbreviated_path` to check `= 'no'` instead of yes?

qatsi:
> What exactly are you trying to accomplish...?

I think it is better to explain with pictures:) In my example when you set it to `auto`, Ruby information is show only in Ruby on Rails project folder (on the second picture `fish` is example Rails project). When you outside of project directory prompt is less informative. Third picture shows custom formatting.

bobthecow:
Gotcha. To be honest, I don't use `rvm` (I use `rbenv`) so the `rvm` version of the ruby version segment is probably less optimal :)

It looks like you're making `rvm` support work similar to the way `rbenv` does. If we make it work exactly the same, I'd prefer just changing it and removing the option altogether.

So what `rbenv` does is suppress the ruby version segment entirely if you're using the "default" version, and only shows it for overrides. [See the implementation here](https://github.com/oh-my-fish/theme-bobthefish/blob/master/fish_prompt.fish#L402). In the `rvm` case, I'd guess this means something like…

In any given directory:

1. If the current ruby interpreter is the same as the global ruby interpreter, suppress showing the interpreter.
2. If the current ruby version is the same as the global ruby version, suppress showing ruby version.
3. If the current gemset is the same as the global gemset, suppress showing the gemset.

That way you'd only see the pieces which are important, and you'd *always* see the pieces which are important.

bobthecow:
The overarching theme here is that bobthefish is opinionated about the best way to do things. That's a lot of the reason people like it. So if we have a strong opinion about something, let's do that by default. If it's enough better than the other options, let's leave off configuration options as well :)

bobthecow:
One of the guiding principles of bobthefish is to only show relevant information. For example, it never shows your user name if it's the default. It never shows your host name unless you're ssh'd into another host. It never shows git information unless it's relevant. In every case, it tries to show you as little information as possible, only exposing things that are different from the "default". So in the case of the `rvm` ruby version segment, I'd say bobthefish is failing. Let's fix it :)

qatsi:
I've finally understood idea and philosophy of this theme:) Thanks for clarification, I'll abide.

As for `rvm` it is very tricky to make it's behavior similar to `rbenv`. It seems `rbenv` is very unobtrusive and frugal. `rvm` is heavy and feature rich, just look at [this](https://github.com/rvm/rvm/blob/master/bin/rvm-prompt#L110:L131) for example :) I'll experiment with rvm segment bit later.

Currently I've fixed mentioned regression
> Please keep the which version :)

and changed title defaults

> I think not displaying the process and using the abbreviated path in the title are better defaults

I'm going to read whole this conversation in order to carefully implement discussed points.

bobthecow:
:+1: :)

Hey, it looks like passing the `--no-default` flag in might get us most of the way there — it looks like it suppresses the whole thing when using the system default ruby? Granted, it's not the most optimal. And it wouldn't suppress default interpreters and such, but it's a start.

qatsi:
>  it looks like it suppresses the whole thing when using the system default ruby?

It works bit different way. Default means just creating alias called `default`, rvm will activate it for each new shell session. So, in simple words `system ruby` != `default ruby` in rvm. This fact leads to some extra difficulties. We need to figure out what is default for user, more precisely it is necessary find default interpreter, version and gemset. After that we should compare what the difference between default and current environment and show it. That is what I'm going to experiment with.

I'd like to avoid using `--no-default` due to it's speed. As I've found `rvm-prompt --no-default i v g` is slowest:
```` 
$ time rvm-prompt i v g
ruby-2.2.3@global
        0.00 real         0.00 user         0.00 sys

$ time rvm-prompt -d i v g
        0.72 real         0.43 user         0.28 sys
````
Finding default with `rvm` is also slow, about half of second:)
````
$ time rvm alias show default
ruby-2.2.3@global
        0.39 real         0.22 user         0.15 sys

````
I've found how `rvm` stores its aliases internally:
````
$ time cat ./rvm/config/alias
default=ruby-2.2.3@global
        0.00 real         0.00 user         0.00 sys
````
May be fastest way it's to parse default alias.

**Comment by qatsi** [comment]
Setting `theme_display_ruby_format` removed as useless. Two functions added `__bobthefish_rvm_parse_ruby` and `__bobthefish_rvm_info`. I have no idea how to implement this without wasting global scope, so I had to create six global variables with `__rvm_` prefix, for example:
````
$__rvm_current_ruby_gemset        (Variable: test)
$__rvm_current_ruby_interpreter  (Variable: jruby)
$__rvm_current_ruby_version     (Variable: 1.7.19)
$__rvm_default_ruby_gemset      (Variable: global)
$__rvm_default_ruby_interpreter   (Variable: ruby)
$__rvm_default_ruby_version      (Variable: 2.2.3)
````
Fortunately, right now they don't conflict with others.

Setting `theme_display_ruby` has two non-default options `no` (as before) and `auto` (new one).
- **By default** it always shows minimal amount of information (I'll cover this later)
- With `auto` it shows minimal amount of information, but only in Ruby-related folders i.e. *it's even more unobtrusive*.
- With `no` it is possible to disable it completely. Forever.

First picture shows *when* it works.



Now about what it shows if it show something:
- If Ruby and gemset default, don't show
- If interpreter differs form default interpreter, show everything: interpreter, version and non-default gemset
- If version differs form default version, show version and non-default gemset
- If gemset differs form default or 'global' gemset, just show it

First picture shows *how* it works.



That is it.

File: README.md [hunk 6]
```diff
@@ -65,16 +61,26 @@ set -g theme_display_git_untracked no
 set -g theme_display_git_ahead_verbose yes
 set -g theme_display_hg yes
 set -g theme_display_virtualenv no
-set -g theme_display_ruby no
+set -g theme_display_ruby no|auto
 set -g theme_display_user yes
 set -g theme_title_display_process yes
 set -g theme_title_display_path no
+set -g theme_title_use_abbreviated_path yes
```
bobthecow: change this to `no` (the rest of them are showing how to override; default here is `yes` so we need to swap this around)

fish_prompt.fish
```diff
@@ -391,10 +388,45 @@ function __bobthefish_prompt_virtualfish -d "Display activated virtual environme
   set_color normal
 end
 
-function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
-  [ "$theme_display_ruby" = 'no' ]; and return
+function __bobthefish_rvm_parse_ruby -d 'Parse RVM Ruby string'
+  # Function arguments:
+  # - 'ruby-2.2.3@rails', 'jruby-1.7.19'...
+  # - 'default' or 'current'
+  set -l __ruby (echo $argv[1] | cut -d '@' -f 1 2>/dev/null)
```
bobthecow: use a named arg instead of `$argv[1]` here.

fish_prompt.fish
bobthecow: I still think this is wrong. If you have this:

```
foo/
foo/.ruby-version
foo/bar/
```

… and you're currently inside `foo/bar/`, your ruby version is affected by the `.ruby-version` in parent. So not showing the custom ruby version when you're inside a subfolder is wrong.

README.md
```diff
@@ -1,12 +1,11 @@
 # bobthefish
 
-bobthefish is a Powerline-style, Git-aware fish theme optimized for awesome.
+`bobthefish` is a Powerline-style, Git-aware [Oh My Fish][omf] theme optimized for awesome.
```
bobthecow: technically, it's not. it's a fish theme. you can install it without omf, and it has no dependencies on omf. that just happens to be the easiest way to get it installed :)

bobthecow:
Sorry, I missed the update. It's starting to come together, thanks for all your work on it :)

fish_prompt.fish
qatsi: Ok. I see. I'll fix that, not sure how to do it better. Just I'm going to traverse all parents of current directory and check how fast is it.

bobthecow:
This looks like it's in a good place other than the `theme_display_ruby=auto` business. If you pulled that out for now, I'd be happy to merge the rest and we can re-address that in another PR?

qatsi:
Hi, I've fixed typos and removed all mentions and logic of `auto`.
It seems it's bit complicated than I expected, so let re-address it for future.
Thanks for advises.

fish_prompt.fish
```diff
@@ -391,10 +388,45 @@ function __bobthefish_prompt_virtualfish -d "Display activated virtual environme
...
+    if [ "$__rvm_current_ruby_gemset" = 'global' ]; rvm-prompt i v
+      else; rvm-prompt i v g; end
+  # If version differs form default version
+  else if [ "$__rvm_default_ruby_version" != "$__rvm_current_ruby_version" ]
+    if [ "$__rvm_current_ruby_gemset" = 'global' ]; rvm-prompt v
+    else; rvm-prompt v g; end
+  # If gemset differs form default or 'global' gemset, just show it
+  else if [ "$__rvm_default_ruby_gemset" != "$__rvm_current_ruby_gemset" ]
+    rvm-prompt g;
+  end
```
bobthecow: Does it make sense to delete all these globals at the end of the function when they're no longer needed?

qatsi: Hi! I think it does. So I've updated PR with 1fd2a4f.
Function become bigger, however I think more important to keep global scope clean.

bobthecow:
Looking good, thanks!

Do you mind squashing the history a bit? It's getting messy. Feel free to mash it all into one commit if you don't feel like splitting it up :)

qatsi:
Done. I've made interactive rebase against master and rewritten history. Everything squashed in one single commit.

> **Do not rebase commits that exist outside your repository.** If you follow that guideline, you’ll be fine. If you don’t, people will hate you, and you’ll be scorned by friends and family.
*[GitBook](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#The-Perils-of-Rebasing)*

bobthecow:
Great, thank you!

Issue #12, opened 2015-10-24
Some git repos trigger very slow prompt display


pik:
Hello - thanks for the comprehensive theme package - I'm having an issue lately with the delay during fish-prompt rendering. 

I modified the fish_prompt function to the following to print the loading time between when the prompt_function was called and when it was resolved:

 ```
set start (gdate +%s%N)
  __bobthefish_prompt_status
  __bobthefish_prompt_user
  __bobthefish_prompt_rubies
  __bobthefish_prompt_virtualfish

  set -l git_root (__bobthefish_git_project_dir)
  set -l hg_root  (__bobthefish_hg_project_dir)
  if [ (echo "$hg_root" | wc -c) -gt (echo "$git_root" | wc -c) ]
    __bobthefish_prompt_hg $hg_root
  else if [ "$git_root" ]
    __bobthefish_prompt_git $git_root
  else
    __bobthefish_prompt_dir
  end
  echo -n -s (math (gdate +%s%N) - $start) | ruby -e 'print $stdin.gets.to_f / 10**9; $stdout.flush'
  __bobthefish_finish_segments
```

Here are some times:
 
```
2.2.3 ~/git 0.096219 cd ipython
 2.2.3 ~/g/ipython   master * 0.155673 cd ../theme-bobthefish/
 2.2.3 ~/g/theme-bobthefish   master * 0.122063 cd ../ruby-pg/
 2.2.3 ~/g/ruby-pg   master 0.136022 cd ../rails                                                                                                  
 2.2.3 ~/g/rails   master $ 0.157983 cd ../tradegecko/                                                                                            
 2.2.3 ~/g/tradegecko reporting-saving $… 0.838621
```

PR #13, opened 2015-10-24
Reduce git and rbenv related rendering bottlenecks

pik:
Fixes https://github.com/oh-my-fish/theme-bobthefish/issues/12

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -393,16 +393,15 @@ end
 
 function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
   [ "$theme_display_ruby" = 'no' ]; and return
-  set -l ruby_version
-  if type rvm-prompt >/dev/null 2>&1
+  if which rvm-prompt >/dev/null 2>&1
     set ruby_version (rvm-prompt i v g)
-  else if type rbenv >/dev/null 2>&1
+  else if which rbenv >/dev/null 2>&1
     set ruby_version (rbenv version-name)
     # Don't show global ruby version...
-    [ "$ruby_version" = (rbenv global) ]; and return
+    set -q RBENV_ROOT; and set rbenv_root $RBENV_ROOT; or set rbenv_root ~/.rbenv
+    [ "$ruby_version" = (cat $rbenv_root/version 2>/dev/null; or echo 'system') ]; and return
   end
   [ -z "$ruby_version" ]; and return
-
   __bobthefish_start_segment $__bobthefish_ruby_red $__bobthefish_lt_grey --bold
   echo -n -s $ruby_version ' '
   set_color normal
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -322,7 +322,7 @@ function __bobthefish_prompt_git -a current_dir -d 'Display the actual git state
   set -l new ''
   set -l show_untracked (git config --bool bash.showUntrackedFiles)
   if [ "$theme_display_git_untracked" != 'no' -a "$show_untracked" != 'false' ]
-    set new (command git ls-files --other --exclude-standard)
+    set new (command git ls-files --other --exclude-standard --directory)
     if [ "$new" ]
       if [ "$theme_avoid_ambiguous_glyphs" = 'yes' ]
         set new '...'


fish_prompt.fish
```diff
@@ -322,7 +322,7 @@ function __bobthefish_prompt_git -a current_dir -d 'Display the actual git state
   set -l new ''
   set -l show_untracked (git config --bool bash.showUntrackedFiles)
   if [ "$theme_display_git_untracked" != 'no' -a "$show_untracked" != 'false' ]
-    set new (command git ls-files --other --exclude-standard)
+    set new (command git status --porcelain)
```
bobthecow: this doesn't actually show when there are new files, though, it shows whenever there are any modifications at all :-/

i haven't seen significant slowdowns using this. i'm curious whether adding `--directory` would speed things up for you?

bobthecow:
did using which vs type make a difference in speed?

pik:
Yes which is faster than type.

fish_prompt.fish
```diff
@@ -393,16 +393,15 @@ end
...
 function __bobthefish_prompt_rubies -d 'Display current Ruby (rvm/rbenv)'
   [ "$theme_display_ruby" = 'no' ]; and return
-  set -l ruby_version
-  if type rvm-prompt >/dev/null 2>&1
+  if which rvm-prompt >/dev/null 2>&1
     set ruby_version (rvm-prompt i v g)
-  else if type rbenv >/dev/null 2>&1
-    set ruby_version (rbenv version-name)
+  else if which rbenv >/dev/null 2>&1
+    set ruby_version (cat .ruby-version 2>/dev/null)
```
pik: This should be the same as rbenv --version behaviour now I think ?

fish_prompt.fish
```diff
@@ -393,16 +393,15 @@ end
...
     # Don't show global ruby version...
-    [ "$ruby_version" = (rbenv global) ]; and return
+    set -q $RBENV_ROOT; and set rbenv_root ~/.rbenv; or set rbenv_root $RBENV_ROOT
```
bobthecow: I think this is backwards: if `$RBENV_ROOT` is defined, use something that's not `$RBENV_ROOT`; if `$RBENV_ROOT` is _not_ defined, use `$RBENV_ROOT` :)

fish_prompt.fish
bobthecow: Unfortunately, using just `.ruby-version` in the current directory won't give an accurate result. To properly emulate `rbenv version-name`, we would have to walk backwards up the current path and look for `.ruby-version` at every level. Do you have a sense for how much of the slowdown was related to this particular line? Is it worth re-implementing that logic in fish? Or would the speed improvements mentioned in the rbenv issue take care of this as of the next release anyway?

fish_prompt.fish
pik: By default I think $RBENV_ROOT is not defined and it looks in `~/.rbenv` no ?

fish_prompt.fish
pik: Yeah your right, that's not going to work. it's about 30ms for rbenv version-name, it's not terrible but if you pay 100ms for the git info and and potentially whatever python virtualenv may cost it will be quite bad.

bobthecow: Yeah, I'm all for improving the actual library where possible, but if we can save all our users 30ms every time they show their prompt in the meantime, I wouldn't complain about that either :)

fish_prompt.fish
bobthecow: :)

bobthecow:
On the `git ls-files` front, I'm curious: did `--directory` give you enough of a speedup? If not, we can mess with parsing the porcelain output of `git status`. But my intuition would be that it couldn't ever be _faster_ to get a complete status than to get a list of the untracked files using a low level call. 

One thing that might be a fairly significant win would be calling `git status --porcelain` once and passing the result into all of these git flag functions to extract info out of.

bobthecow:
A bit more concrete info for that last gut level statement :)

On a fairly large repo I use, on a Mac OS X system (which is much slower with large git repos than linux filesystems), I got this for our current flag operations when the local copy is in an unchanged state (i.e. nothing to report):

```
dirty     7ms
staged    3ms
stashed   1ms
untracked 3ms
----
total     14ms
```

To put that in context just running `git status --porcelain` once in that repo takes 24ms, before doing any of the parsing. So at least for this one repo, in an unchanged state, in the best case possible `git status` cant touch the individual calculations.

That's not saying the individual calculations can't be improved. I imagine `--directory` is a huge improvement for repos with entire untracked subdirectories.

bobthecow:
Merging as-is, we can continue iterating if we find further improvements :)

pik:
Thanks for keeping up with this. The `--directory` flag was the really huge one - here is the difference for me (also on OS X atm): 
 
```
> timeit "git ls-files --other --exclude-standard --directory" 10                                                                            
0.194094
>timeit "git ls-files --other --exclude-standard" 10                                                                                        
3.467714
```

Also I'll try and look at rbenv to see if some more time can be won there.

PR #15, opened 2015-10-29
Support vi mode

oschrenk:
This is initial pull request to support displaying the vi-mode in fish. Would love to get some feedback, especially on the color choices.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -370,22 +370,19 @@ function __bobthefish_prompt_dir -d 'Display a shortened form of the current dir
 end
 
 function __bobthefish_prompt_vi -d 'Display vi mode'
-  if [ "$theme_display_vi" = 'yes' ]
-    if [ "$fish_bind_mode" != "$theme_display_vi_hide_mode" ]
-      switch $fish_bind_mode
-      case default
-        __bobthefish_start_segment $__bobthefish_med_grey $__bobthefish_dk_grey --bold
-        echo -n -s 'N '
-      case insert
-        __bobthefish_start_segment $__bobthefish_lt_green $__bobthefish_dk_grey --bold
-        echo -n -s 'I '
-      case visual
-        __bobthefish_start_segment $__bobthefish_lt_orange $__bobthefish_dk_grey --bold
-        echo -n -s 'V '
-      end
-      set_color normal
-    end
+  [ "$theme_display_vi" = 'yes' -a "$fish_bind_mode" != "$theme_display_vi_hide_mode" ]; or return
+  switch $fish_bind_mode
+    case default
+      __bobthefish_start_segment $__bobthefish_med_grey $__bobthefish_dk_grey --bold
+      echo -n -s 'N '
+    case insert
+      __bobthefish_start_segment $__bobthefish_lt_green $__bobthefish_dk_grey --bold
+      echo -n -s 'I '
+    case visual
+      __bobthefish_start_segment $__bobthefish_lt_orange $__bobthefish_dk_grey --bold
+      echo -n -s 'V '
   end
+  set_color normal
 end
 
 function __bobthefish_virtualenv_python_version -d 'Get current python version'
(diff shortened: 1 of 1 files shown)


fish_prompt.fish
```diff
@@ -367,6 +368,23 @@ function __bobthefish_prompt_dir -d 'Display a shortened form of the current dir
   __bobthefish_path_segment "$PWD"
 end
 
+function __bobthefish_prompt_vi -d 'Display vi mode'
+  [ "$theme_display_vi" = 'no' ]; and return
+  set -l vi_mode
+  switch $fish_bind_mode
+  case default
+    __bobthefish_start_segment $__bobthefish_med_grey $__bobthefish_dk_grey --bold
+    echo -n -s 'N'
```
bobthecow: Default probably shouldn't show anything at all … especially as most users won't use vi mode at all, so adding a segment to their prompt which never changes and adds nothing is kind of a crappy thing to do :)

But even for the users who do, it probably shouldn't show in normal mode. bobthefish goes to a lot of effort to show you only the things which are important. For example, there's no "last exit code was fine" flag, but there is a "last exit code was an error". There's no "currently up to date" git flag, but there is one for each possible permutation of ahead, behind, stashed, dirty, untracked, etc.

fish_prompt.fish
```diff
@@ -367,6 +368,23 @@ function __bobthefish_prompt_dir -d 'Display a shortened form of the current dir
   __bobthefish_path_segment "$PWD"
 end
 
+function __bobthefish_prompt_vi -d 'Display vi mode'
+  [ "$theme_display_vi" = 'no' ]; and return
+  set -l vi_mode
```
bobthecow: looks like this is unused?

fish_prompt.fish
```diff
@@ -427,5 +445,6 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   else
     __bobthefish_prompt_dir
   end
+  __bobthefish_prompt_vi
   __bobthefish_finish_segments
```
bobthecow: it seems like it would be more consistent with the way powerline works, for example, to move the vi segment all the way to the left. that would also separate it from the git segment so you wouldn't have back-to-back green segments when in insert mode inside an up to date git repo.

that said, i don't use vi mode, so i don't have a strong sense of where would make the most sense to put it.

bobthecow:
Oh, there should probably be a space at the end of those segments. It looks cramped if you have a letter right before the closing arrow.

SolarAquarion:
look at budspencer for example for how you can do vi mode notifications within the powerline looks

fish_prompt.fish
oschrenk: Removed

fish_prompt.fish
oschrenk: People have a very different opinion of what the default mode is though. I always start in insert mode and not in normal mode for example.

vim-airline always shows mode, regardless which mode that is. I added an option to hide a single mode, but that would still mean that the user has to add some configuration if he wants to hide it.

oschrenk:
- moved vi segment to front (just after a failed status)
- user has to add configuration to actually enable segment
- added whitespace to not look so cramped
- reversed logic for integration, user has to specfically `set -g theme_display_vi yes`

fish_prompt.fish
```diff
@@ -367,6 +369,25 @@ function __bobthefish_prompt_dir -d 'Display a shortened form of the current dir
   __bobthefish_path_segment "$PWD"
 end
 
+function __bobthefish_prompt_vi -d 'Display vi mode'
+  if [ "$theme_display_vi" = 'yes' ]
```
bobthecow: to be a bit more consistent with the other segments, let's do an early return rather than an if block that wraps the whole function.

```fish
[ "$theme_display_vi" != 'yes' ]; and return
```

bobthecow:
Do you mind posting a screenshot?

This will also need a rebase against master so I can merge it.

oschrenk:
Rebased. Tell me if you want me to squash the commits.

vi-mode

bobthecow:
Thanks! No need to squash.

PR #19, opened 2016-01-02
Add missing `-g` flag to __bobthefish_* global variables

### LINKED ISSUE #18: fundle compatibility
The theme is displayed incorrectly when installed with [fundle](https://github.com/tuvistavie/fundle):

```fish
fundle plugin 'oh-my-fish/theme-bobthefish'
```

screenshot - 12172015 - 11 42 18 am

It can be fixed by moving all fish sources to `functions` subfolder.

bucaran:
White flag here and sincere PR to fix a minor issue.

Basically you want to add missing the `-g` flag to __bobthefish_* global variables.

For three reasons:

1. Because these are global variables, and `-g` shows this intent clearly.

2. For consistency with `__bobthefish_current_bg`. 

3. Because it breaks the prompt if you are sourcing the bobthefish files without `exec fish`.

If you source `fish_prompt.fish` from inside a function, these variables will be scoped to _that_ function, thus making them private and causing the prompt to display incorrect default colors.

I remember having this problem earlier when I was first prototyping with Wahoo, and the only solution I found at the time was to `exec fish` which is what OMF does as well.

You have expressed your dissatisfaction to using this method before:
* https://github.com/oh-my-fish/oh-my-fish/pull/88#issuecomment-144879571

Quoting you:
> i just think we need to stop replacing the shell. it makes everything awful and does (almost) no good.

Also see here:
* https://github.com/oh-my-fish/oh-my-fish/issues/189

While this PR does not solve the problem upstream, it paves the way for a future `source` builtin based solution.

Also closes #18.

Regards


bobthecow:
I'd actually prefer to make all of these local variables, but since fish doesn't support private functions they're script local. `__bobthefish_current_bg` is intentionally global, the rest are more of a constant than anything else, so I'd strongly prefer them not to leak. I don't know if there's a better way to handle it, though, so I'm open to suggestions (or to using globals, if that's really the only way).

fishfreedom:
While the sentiment is legit, the fact is these variables are global regardless of one's preferences, because that's just how fish works. By adding the -g you express your clear intent and also add compatibility to fundle and other systems without breaking OMF along the way.

Not only we want a more solid private scope in fish, we want private function scope! But the point is, these variables are globals. Please consider merging this PR.

bpinto:
I don't think we should use `-g`. This intention of these variables are not to be global.

```
function inner_function -S
  echo 'abc'
  echo $outter
end

function outter_function
  set -l outter out
  inner_function
end

$ utter_function
abc
out
```

bobthecow:
@bpinto Yep, that works (at least for fishline). Thanks!

PR #20, opened 2016-02-01
Add <stopping> to Vagrant state indicator

sashaegorov:
Hi,
Today I've observed weird state `<stopping>` in my prompt, so I've added related replacement.
Unfortunately I can't reproduce this, because in most cases VirtualBox changes VM's state from `<running >` to `<poweroff>` rapidly.


bobthecow:
Heh :)

PR #25, opened 2016-03-03
Add show error code option

paxy97:
I like my prompt to show the actual error code, so I added an option to show it. It can be toggled via the `$show_error_code` variable. The default is to just show an exclamation mark as before


bobthecow:
Name it `$theme_show_error_code` and add it to the README and you've got yourself a deal :)

bobthecow:
Maybe `theme_show_exit_status` would be a better name?

paxy97:
I'm not completely sure on how pull requests work, but I think it should be updated. Also I added it to the readme in the option listing and at the end, hopefully that's the right place

bobthecow:
It looks good. Thank you!

PR #28, opened 2016-03-19
sometimes you do not want a right prompt

JerrySievert:
tiny pull request, but sometimes it's nice to not have a right side prompt - this allows for disabling the helpful right hand prompt, but still keep the awesome left-hand prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ set -g theme_avoid_ambiguous_glyphs yes
 set -g theme_nerd_fonts yes
 set -g theme_show_exit_status yes
 set -g default_user your_normal_user
+set -g theme_display_right no
 ```
 
 **Title options**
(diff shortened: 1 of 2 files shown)


bobthecow:
I feel like the fishier way would be to define your own `fish_right_prompt`, `fish_greeting`, etc if you don't want the ones the theme provides. In your case, putting `function fish_right_prompt; end` in `~/.config/omf/init.fish` would prevent bobthefish right prompt from even running.

then you could even put something you like more in the right prompt ;)

JerrySievert:
i'm not actually using omf - it clobbered my whole environment, and i got very mad at it.

i've installed bobthefish with `fisher`, which does not clobber my environment.

bobthecow:
I'm sorry to hear about that. Did you open an issue with OMF so they're aware of your problem?

I haven't used fisherman, but the same answer applies to just about any fish package manager. Add a higher priority version of the same function name and it will override the theme default.

JerrySievert:
seems to be by design for OMF, as it purposely backs up and overwrites your config.fish

bobthecow:
Closing this, as it's an issue with the package manager, not the theme, if user defined functions don't override theme functions.

PR #30, opened 2016-03-29
Reset --bold as well after long long running command

JelteF:
If this is not done the `--bold` flag is not reset and the command that is typed next will be bold as well:
bold_command


bobthecow:
:+1:

PR #40, opened 2016-05-15
Fix right prompt colour when $fish_color_autosuggestion is bold

### LINKED ISSUE #39: Right prompt uses incorrect colour if $fish_color_autosuggestion is bold
The colour of the right prompt is based on `$fish_color_autosuggestion`, which is great.
However, if `$fish_color_autosuggestion` has options such as bold, this fails and the default text colour is used. (This happens with colourschemes such as Solarized, where the bold colours are not in fact really bold.)

This is because you are doing `set_color $fish_color_autosuggestion[1]`.
If `$fish_color_autosuggestion` is, for example, "--bold green", this fails because `$fish_color_autosuggestion[1]` is just "--bold".

Instead, you should be using `set_color $fish_color_autosuggestion` (ie, without the `[1]`).

Is the `[1]` there for a reason? Removing it doesn't seem to break anything.

I will submit a pull request later.

(Sorry for three issues at once! :-) )

caesar:
Fixes #39

### DIFF
--- a/fish_right_prompt.fish
+++ b/fish_right_prompt.fish
@@ -20,7 +20,7 @@ function __bobthefish_cmd_duration -S -d 'Show command duration'
   end
 
   set_color $fish_color_normal
-  set_color $fish_color_autosuggestion[1]
+  set_color $fish_color_autosuggestion
   echo -n $__bobthefish_left_arrow_glyph
 end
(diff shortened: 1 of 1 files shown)


bobthecow:
👍

Issue #37, opened 2016-05-15
Option not to flag git stashes


caesar:
I have a lot of git stashes in some of my projects. Often I'll save half-finished work in a stash while working on something else.

It's really annoying to have bobthefish constantly show an orange git status line with the $ indicator. I'd like it to show as clean when there are no unchanged files in the current branch.

Would it be possible to add an option to disable this behaviour?

Happy to submit a pull request if you will accept a new option. :-)

bobthecow:
Let's close this for now, because #43. We can re-address later if you feel that's not enough.

caesar:
Thanks, I think #43 solves this for me.

On May 22 2016, at 8:27 pm, Justin Hileman &lt;notifications@github.com&gt;
wrote:  

> Closed [#37](https://github.com/oh-my-fish/theme-bobthefish/issues/37).

>

> —  
You are receiving this because you authored the thread.  
Reply to this email directly or [view it on GitHub](https://github.com/oh-my-
fish/theme-bobthefish/issues/37#event-668370232)![](https://github.com/notific
ations/beacon/ABfnNFw1ZivhsEyVb657PSCWEJceExegks5qEPR2gaJpZM4Ie8JY.gif)

bobthecow:
Great!

PR #41, opened 2016-05-15
Add option to disable git stash flagging

caesar:
Adds the option requested in #37.

If you prefer not to add new options I understand, but this would fix my main annoyance with bobthefish so I'd love it if it could get in! :-)

As you can see I have used the same mechanism that you had used with the `$theme_display_git_untracked` option.

Any comments or changes you would like, let me know.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -19,6 +19,7 @@
 #
 #     set -g theme_display_git no
 #     set -g theme_display_git_untracked no
+#     set -g theme_display_git_stashed no
 #     set -g theme_display_git_ahead_verbose yes
 #     set -g theme_git_worktree_support yes
 #     set -g theme_display_vagrant yes
@@ -382,8 +383,12 @@ end
 function __bobthefish_prompt_git -S -a current_dir -d 'Display the actual git state'
   set -l dirty   (command git diff --no-ext-diff --quiet --exit-code; or echo -n '*')
   set -l staged  (command git diff --cached --no-ext-diff --quiet --exit-code; or echo -n '~')
-  set -l stashed (command git rev-parse --verify --quiet refs/stash >/dev/null; and echo -n '$')
   set -l ahead   (__bobthefish_git_ahead)
+
+  set -l stashed ''
+  if [ "$theme_display_git_stashed" != 'no' ]
+    set stashed (command git rev-parse --verify --quiet refs/stash >/dev/null; and echo -n '$')
+  end
 
   set -l new ''
   set -l show_untracked (command git config --bool bash.showUntrackedFiles)


bobthecow:
The reason `$theme_display_git_untracked` exists is entirely performance related (see oh-my-fish/oh-my-fish-legacy#434).

In general, I prefer not to add too many options… bobthefish is opinionated about what makes a good prompt, and that's part of the reason people think it's a good prompt :)

So is there a way we can solve your main annoyance whilst still being opinionated about it? What if, instead of an option to disable it, there were an option to make it less annoying? Maybe just the `$` but not change the color to orange?

bobthecow:
In fact, almost all of the options exist for performance reasons. If you never use {git,hg,worktrees,vagrant,virtualenv,ruby,vi mode}, then disabling them makes everything a _lot_ faster. A handful of the options are opt-in and disabled by default — hg, vagrant, git worktrees, etc — because they're less common and/or a lot slower to enable by default. Otherwise, bobthefish tries to do things "right" rather than "configurable" :P

caesar:
Yep, I understand that and do see the value in being opinionated rather than giving too many options.

Though I would prefer personally to see no indication of the existence of stashes, I could live with showing the $ without not changing the colour; if you feel that would be a good solution.

As a matter of fact, incidentally, checking for stashes _does_ of course have a performance cost, even if it's not huge.

caesar:
As a matter of fact all the colours seem a bit odd to me. I'd say the time the colour _should_ change, but doesn't, is when you have unstaged changes.

I'd go
- unstaged changes -> red
- staged changes -> yellow/orange
- commited -> green

Or does this not make sense?

(I realise this isn't really directly related to the current pull request. Sorry.)

bobthecow:
Unstaged changes are red. Staged changes are _also_ red, though, and I could see a case for changing the color on them to orange or something :)

I do think stashed is important, but maybe not color-change-important. What if it was these flags:

* untracked: `...`
* unstaged: `*`
* staged: `~`
* stashed: `$`

… and these colors:

* anything unstaged: `red`
* else if anything staged: `orange`
* else: `green`

?

bobthecow:
> (I realise this isn't really directly related to the current pull request. Sorry.)

This is directly related. It's us working out how to solve the intent of this pull request, even if it's not with the code from this pull request :)

caesar:
Yes, that makes sense to me.

I was going to comment that I am not seeing a colour change for unstaged changes, but then I realised that the I was really talking about _untracked_ files. I hadn't really thought of this as a different case to (other) unstaged changes.
So that raises the question, should untracked files also cause the prompt to go orange? It feels more natural to me, but I don't know if it's actually better or not now I think about it.

For reference this screenshot seems to show all the variations at the moment, other than stashes.
screen shot 2016-05-17 at 13 18 31

bobthecow:
Untracked files intentionally don't change the color. I feel like a good rule of thumb for a color change is "if I do `git reset --hard` will I lose anything?

caesar:
Okay, maybe that makes sense. I don't have a strong opinion on that one.

Just sent a PR for the new behaviour discussed here. Let's continue the discussion at #43.

bobthecow:
sgtm

Issue #38, opened 2016-05-15
Option not to require Powerline fonts


caesar:
It would be really nice if there was an option to disable the use of Powerline glyphs. I love all the features of bobthefish but would sometimes prefer not to have to use a special font.

In powerline for Vim, for example, it is possible to disable the use of Powerline glyphs (It just shows square ends instead of arrows, and doesn't show the git branch icon, for example).

This would also save a few characters on small screens, I guess...

I don't know how much work it would be to go through making each use of a powerline glyph depend on a central option. But would you accept a pull request? Or do you prefer not to add new options...?

PR #42, opened 2016-05-15
Add option to not use Powerline glyphs

caesar:
Adds the option requested in #38.

Seems to be quite simple, though it's possible I've missed some.

I hope this will be accepted as it would be great for compatibility at very little cost.

Let me know if you'd like me to do anything differently.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -629,7 +629,7 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   set -l __bobthefish_left_arrow_glyph        \uE0B3
 
   if [ "$theme_powerline_fonts" = "no" ]
-    set __bobthefish_branch_glyph            ''
+    set __bobthefish_branch_glyph            '\u2387'
     set __bobthefish_ln_glyph                ''
     set __bobthefish_padlock_glyph           ''
     set __bobthefish_right_black_arrow_glyph ''
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -629,7 +629,7 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   set -l __bobthefish_left_arrow_glyph        \uE0B3
 
   if [ "$theme_powerline_fonts" = "no" ]
-    set __bobthefish_branch_glyph            '\u2387'
+    set __bobthefish_branch_glyph            \u2387
     set __bobthefish_ln_glyph                ''
     set __bobthefish_padlock_glyph           ''
     set __bobthefish_right_black_arrow_glyph ''
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -620,22 +620,22 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   set -l last_status $status
 
   # Powerline glyphs
-  set -l __bobthefish_branch_glyph            ''
-  set -l __bobthefish_ln_glyph                ''
-  set -l __bobthefish_padlock_glyph           ''
-  set -l __bobthefish_right_black_arrow_glyph ''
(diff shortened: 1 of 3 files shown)


fish_prompt.fish
```diff
@@ -619,13 +620,23 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
...
-  set -l __bobthefish_left_arrow_glyph        \uE0B3
+  set -l __bobthefish_branch_glyph            ''
+  set -l __bobthefish_ln_glyph                ''
+  set -l __bobthefish_padlock_glyph           ''
+  set -l __bobthefish_right_black_arrow_glyph ''
+  set -l __bobthefish_right_arrow_glyph       ''
+  set -l __bobthefish_left_black_arrow_glyph  ''
+  set -l __bobthefish_left_arrow_glyph        ''
+
+  if [ "$theme_powerline_fonts" != "no" ]
```
derekstavis: I would prefer making powerline fonts the default, as this would have less performance penalties on the prompt at every render.

caesar: @derekstavis Sure, makes sense. Done.

bobthecow: 👍

fish_prompt.fish
```diff
@@ -627,6 +628,16 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   set -l __bobthefish_left_black_arrow_glyph  \uE0B2
   set -l __bobthefish_left_arrow_glyph        \uE0B3
 
+  if [ "$theme_powerline_fonts" = "no" ]
+    set __bobthefish_branch_glyph            ''
```
bobthecow: we don't use most of the non-arrow glyphs, but we do use branch. is there a good unicode character  that approximates this?

bobthecow:
What should happen if someone disables powerline fonts but enables nerd fonts?

fish_prompt.fish
caesar: What about ⎇ (U+2387, ALTERNATIVE KEY SYMBOL)?
However it's not present in all fonts, I believe.
(As a matter of fact I don't think it's important to show the branch icon in 'fallback' mode. It is fairly clear what the branch name means anyway.)

caesar:
Interesting. Why would they do that? :-)
I think nerd fonts include powerline anyway don't they?

bobthecow:
I'm not saying it's sane, just that it's possible given the set of theme configuration variables :)

fish_prompt.fish
bobthecow: Let's start with ⎇ and refine later if necessary?

caesar:
Added ⎇ symbol.

As for disabling powerline but enabling nerd fonts, though it would be an odd thing to do, I think the current behaviour matches expectations. It enables nerd fonts (other than the powerline glyphs), but disables powerline.
The alternative would be for the nerd option to override the powerline one, but I don't think that would match expectations.
That said, this is all a hypothetical edge case anyway because there seems to be no valid reason to both enable nerd and disable powerline.

fish_prompt.fish
```diff
@@ -627,6 +628,16 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   set -l __bobthefish_left_black_arrow_glyph  \uE0B2
   set -l __bobthefish_left_arrow_glyph        \uE0B3
 
+  if [ "$theme_powerline_fonts" = "no" ]
+    set __bobthefish_branch_glyph            '\u2387'
```
bobthecow: let's do this without the quotes to match the rest of 'em.

otherwise, :shipit: !

bobthecow:
Oh, please add the config option to the bottom of README as well.

caesar:
👍

PR #45, opened 2016-05-23
Support chruby ruby_version for show_ruby()

pik:
I've switched to `chruby` (https://github.com/postmodern/chruby) from `rbenv` recently due to the difficulties rbenv shims tend to introduce when debugging ruby runtimes. It's quite a popular alternative so it seems like it would be worth supporting. 

(RUBY_VERSION is blank if chruby is using the system version)


bobthecow:
Thanks! 👍 

Do you mind adding this to the line in `README.md` that mentions ruby versions as well?

bobthecow:
Never mind, I'll do it. Thanks for the commit!

PR #47, opened 2016-06-06
Add configurable color schemes

### LINKED ISSUE #7: Use terminal colors or allow customization
Right now the prompt colors are great but only if they match the color scheme currently used by the terminal.  If you switch to, say, zenburn...things don't match very well.  

One option is for the user to copy the entire theme and replace the hard coded colors with different colors.  But what do you think about either adding an option that would use the 16 colors used by the terminal, or adding an option that lets the user override each of the hard coded colors?

ohspite:
* Refactor fish_prompt.fish to use semantically named variables
  throughout the script rather than specific colors. Set those variables
  to various colors depending on the `theme_color_scheme` variable.

* Add the `__bobthefish_display_colors` function to easily test and
  preview different color schemes.

* Update documentation.

Fixes #7.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -608,15 +608,15 @@ end
 # ===========================
 
 function __bobthefish_display_colors -d 'Print example prompts using the current color scheme'
-  set -g __bobthefish_display_colors 1
+  set -g __bobthefish_display_colors
 end
 
 function __bobthefish_maybe_display_colors -S
-  if test -z $__bobthefish_display_colors
+  if not set -q __bobthefish_display_colors
     return
   end
 
-  set -g __bobthefish_display_colors
+  set -e __bobthefish_display_colors
 
   echo
   set_color normal
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -763,31 +763,41 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
       # Do not set any variables in this section.
 
       # If you want to create your own color scheme, set `theme_color_scheme` to
-      # `user` and define the following variables in your fish startup file. Use
-      # `__bobthefish_display_colors` to easily see what these variables are
-      # used for.
+      # `user` and define the `__color_*` variables listed below in your fish
+      # startup file (`$OMF_CONFIG/init.fish`, or similar).
 
       # The value for each variable is an argument to pass to `set_color -b`.
-      # See See the color schemes below for examples.
+      # You can copy the commented code below as a base for your custom colors.
+      # Use `__bobthefish_display_colors` at the command line to easily see what
+      # these variables are used for.
+
+      # See the built-in color schemes below for more examples.
+
+      # # Example bobthefish color scheme:
+      # set -g theme_color_scheme user
       #
-      # - `__color_initial_segment_exit`
-      # - `__color_initial_segment_su`
-      # - `__color_initial_segment_jobs`
-      # - `__color_path`
-      # - `__color_path_basename`
-      # - `__color_path_nowrite`
-      # - `__color_path_nowrite_basename`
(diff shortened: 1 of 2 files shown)


ohspite:
Feel free to change anything or ask me to change anything. It's pretty straightforward, but in particular I had to make a lot of choices for variables names and you might prefer something different.

All of the user-facing changes should be described in the README, so I won't duplicate those here.  But getting a feel for this should be as easy as

```
> set theme_color_scheme dark
> __bobthefish_display_colors
> set theme_color_scheme solarized
> __bobthefish_display_colors
> set theme_color_scheme base16
> __bobthefish_display_colors
> # etc.
```

fish_prompt.fish
```diff
@@ -614,6 +604,110 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
 end
 
 # ===========================
+# Debugging functions
+# ===========================
+
+function __bobthefish_display_colors -d 'Print example prompts using the current color scheme'
+  set -g __bobthefish_display_colors 1
+end
```
bobthecow: i don't love this, but i don't see an easy way of implementing it otherwise…

fish_prompt.fish
```diff
@@ -614,6 +604,110 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
 end
 
 # ===========================
+# Debugging functions
+# ===========================
+
+function __bobthefish_display_colors -d 'Print example prompts using the current color scheme'
+  set -g __bobthefish_display_colors 1
```
bobthecow: use `set -q`

fish_prompt.fish
```diff
@@ -614,6 +604,110 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+end
+
+function __bobthefish_maybe_display_colors -S
+  if test -z $__bobthefish_display_colors
+    return
+  end
+
+  set -g __bobthefish_display_colors
```
bobthecow: use `set -e __bobthefish_display_colors`

fish_prompt.fish
```diff
@@ -614,6 +604,110 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
 # ===========================
+# Debugging functions
+# ===========================
+
+function __bobthefish_display_colors -d 'Print example prompts using the current color scheme'
+  set -g __bobthefish_display_colors 1
+end
+
+function __bobthefish_maybe_display_colors -S
+  if test -z $__bobthefish_display_colors
```
bobthecow: use `set -q __bobthefish_display_colors`

fish_prompt.fish
ohspite: yes, same here.

ohspite:
Thanks!  Good call on the `set` stuff.

fish_prompt.fish
```diff
@@ -663,30 +757,418 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
...
+
+      # If you want to create your own color scheme, set `theme_color_scheme` to
+      # `user` and define the following variables in your fish startup file. Use
+      # `__bobthefish_display_colors` to easily see what these variables are
+      # used for.
+
+      # The value for each variable is an argument to pass to `set_color -b`.
+      # See See the color schemes below for examples.
+      #
+      # - `__color_initial_segment_exit`
```
bobthecow: should we pre-populate a (commented out) block of `set` commands that can just be copied and pasted into config — use the default dark colors — to make it easier for people who want to customize?

ohspite: Check out the latest commit and let me know what you think.

bobthecow:
Merged. Thanks again for your contribution!

PR #48, opened 2016-06-07
Docker machine display and conditional date

feinstruktur:
I probably shouldn't have rolled this into one PR but those were things I tweaked and I figured I might as well clean it up a little and open a PR :)

What is does is add a switch to be able to switch off the date display:
```
set -g theme_display_date no
```

And it shows which docker machine you're talking to:
screen shot 2016-06-07 at 08 44 36

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,12 +68,14 @@ set -g theme_display_git_untracked no
 set -g theme_display_git_ahead_verbose yes
 set -g theme_git_worktree_support yes
 set -g theme_display_vagrant yes
+set -g theme_display_docker_machine no
 set -g theme_display_hg yes
 set -g theme_display_virtualenv no
 set -g theme_display_ruby no
 set -g theme_display_user yes
 set -g theme_display_vi yes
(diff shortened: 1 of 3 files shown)


bobthecow:
Thanks for your contribution!

The `$__bobthefish_left_arrow_glyph` after the timestamp should be suppressed if `$theme_display_date = no`. That and a `-d` and I think we're good to merge?

feinstruktur:
Right, I thought that glyph was part of the command duration. It’s probably best then I’ll move that to the start of `__bobthefish_timestamp` to keep things together, right?

bobthecow:
Sorry, I guess I wasn't super clear. That glyph is a separator, and should only show if _both_ segments are there. I'll fix it though. Thanks for your contribution :)

bobthecow:
Okay, updated. Let me know if anything's weird for you.

feinstruktur:
Ah, I see what you mean now. Thanks for merging 👍

PR #51, opened 2016-07-15
Add option to hide cmd duration

zaclimon:
In case there are people who want to keep their terminal "clean".

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ set -g theme_display_user yes
 set -g theme_display_vi yes
 set -g theme_display_vi_hide_mode default
 set -g theme_display_date no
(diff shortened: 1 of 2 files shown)


bobthecow:
To disable the greeting, the fishier thing to do is write your own `fish_greeting` function.

zaclimon:
Ohh okk then. is the CMD duration option fishy enough?

bobthecow:
Yeah, I think so :)

Let's change it to `theme_display_cmd_duration`, and add it to the list in the readme?

zaclimon:
This should look fine now. :)

bobthecow:
Great, thanks!

PR #55, opened 2016-08-21
Display git tags instead of short hash when in deteached head state.

rickard-von-essen:
Use `git describe --tags` to get a short description of the commit
relative the nearest tag. Only the tag name will be shown if your
currently on a tag.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -20,6 +20,7 @@
 #     set -g theme_display_git no
 #     set -g theme_display_git_untracked no
 #     set -g theme_display_git_ahead_verbose yes
+#     set -g theme_display_git_tag yes
 #     set -g theme_git_worktree_support yes
 #     set -g theme_display_vagrant yes
 #     set -g theme_display_docker_machine no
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Example screenshot
screen shot 2016-08-21 at 21 21 29

fish_prompt.fish
```diff
@@ -53,7 +54,12 @@
 function __bobthefish_git_branch -S -d 'Get the current git branch (or commitish)'
   set -l ref (command git symbolic-ref HEAD ^/dev/null)
   if [ $status -gt 0 ]
-    set -l branch (command git show-ref --head -s --abbrev | head -n1 ^/dev/null)
+    set -l branch
+    if [ "$theme_display_git_tag" = 'yes' ]
+      set branch (command git describe --tags ^/dev/null)
```
bobthecow: this probably needs an `--always` in there to keep it from falling over if you don't have a tag in your repo yet.

bobthecow:
I've been considering changing the detached head "branch name" command to this:

```
command git describe --tags --exact-match ^/dev/null;
  or command git show-ref --head -s --abbrev | head -n1 ^/dev/null
```

Which would be very similar to yours, except it would show only the hash when HEAD isn't a tag itself, while yours would do something like `v0.1.4-1-g4f7d5e7`. Would having the "tag-or-hash but not both" version be enough for you, or are you actually looking for the extended "tag-plus-hash-plus-more" information?

rickard-von-essen:
That would be great. Tags are most important.

bobthecow:
This is fixed in 4669e3dad6ad

Let me know if you have any issues :)

rickard-von-essen:
Works nicely!

Issue #53, opened 2016-08-15
Display Vagrant status for Parallels Desktop for Mac


rickard-von-essen:
Show Vagrant status for [Parallels Desktop for Mac](https://github.com/Parallels/vagrant-parallels)

rickard-von-essen:
I'll PR this soon.

bobthecow:
:+1:

PR #56, opened 2016-08-21
Improved Vagrant provider detection and support for parallels provider.

rickard-von-essen:
Improved the detection of Vagrant provider. The previous didn't
correctly detect which provider was used.

Added support for detecting the state of Parallels vagrant provider.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -238,7 +238,7 @@ end
 
 function __bobthefish_prompt_vagrant -S -d 'Display Vagrant status'
   [ "$theme_display_vagrant" = 'yes' -a -f Vagrantfile ]; or return
-  set provider (__bobthefish_vagrant_provider)
+  set -l provider (__bobthefish_vagrant_provider)
   switch "$provider"
     case 'virtualbox'
       __bobthefish_prompt_vagrant_vbox
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -238,15 +238,24 @@ end
 
 function __bobthefish_prompt_vagrant -S -d 'Display Vagrant status'
   [ "$theme_display_vagrant" = 'yes' -a -f Vagrantfile ]; or return
-  if type -q VBoxManage
-    __bobthefish_prompt_vagrant_vbox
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Closes #53

fish_prompt.fish
```diff
@@ -238,23 +238,32 @@ end
 
 function __bobthefish_prompt_vagrant -S -d 'Display Vagrant status'
   [ "$theme_display_vagrant" = 'yes' -a -f Vagrantfile ]; or return
-  if type -q VBoxManage
-    __bobthefish_prompt_vagrant_vbox
-  else if grep vmware_fusion Vagrantfile >/dev/null ^&1
-    __bobthefish_prompt_vagrant_vmware
+  set provider (__bobthefish_vagrant_provider)
```
bobthecow: Let's use `set -l` here.

fish_prompt.fish
```diff
@@ -238,23 +238,32 @@ end
...
+    case 'vmware_fusion'
+      __bobthefish_prompt_vagrant_vmware
+    case 'parallels'
+      __bobthefish_prompt_vagrant_parallels
   end
 end
 
-function __bobthefish_vagrant_ids -S -d 'List Vagrant machine ids'
-  for file in .vagrant/machines/**/id
+function __bobthefish_vagrant_provider -S -d 'Get (first) Vagrant provider'
```
bobthecow: To be honest, I only use one Vagrant provider, so I've never tested how it works when you have multiple. Is it possible for there to be VMs from two or more Vagrant providers in the same directory? If so, is there any reason you'd care about all of them?

rickard-von-essen:
I have improved this some more. I'll supply some more code later today.

rickard-von-essen:
Now it should work for all mixed provider cases.

bobthecow:
Looking good! It should be optimized a bit for performance, but I'll add that in a separate commit.

bobthecow:
I optimized it a bit in 8898658 and was able to speed up the time spent on bobthefish's by 10-15x :)

Thanks again for implementing this!

PR #57, opened 2016-08-22
Add Gruvbox Theme

Jguer:
Based on [Gruvbox](https://github.com/morhetz/gruvbox-contrib) by morhetz and following the style of the airline theme "Gruvbox".
screenshot from 2016-08-22 02-52-23

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -1152,6 +1152,41 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
       set __color_rvm                      $ruby_red $grey[1] --bold
       set __color_virtualfish              $blue[2] $grey[1] --bold
 
+    case 'gruvbox'
+      #               light  medium  dark  darkest
+      #               ------ ------ ------ -------
+      set -l red      fb4934 cc241d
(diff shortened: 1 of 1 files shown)


bobthecow:
Looking good!

It would be a bit more true to both bobthefish and gruvbox to use `bg1` or `bg2` for the project path segment (with `fg3` maybe?).

Jguer:
Probably, I'm going to try it. I was using path with the same color as vi_mode_default but then decided to try vi_mode_insert (since I am using emacs mode and by default everything is insert).

Will come back with results.

Jguer:
bg1 with fg2 keeps gruvbox's washed out appearance well

bg1 - fg2
screenshot from 2016-08-22 12-02-25

Added new bg from gruvbox pallete (3c3836) that sits between bg1 and bg2 paired with fg3
screenshot from 2016-08-22 12-11-55

Airline for comparison
screenshot from 2016-08-22 12-14-06

Jguer:
bg1-fg2 (current commit) with the correct gruvbox background
save

bobthecow:
That looks a lot better, thanks!

PR #58, opened 2016-08-26
Added option to disable right prompt

rickard-von-essen:
By setting `set -g $theme_right_prompt = no` the right prompt will be hidden.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ set -g theme_nerd_fonts yes
 set -g theme_show_exit_status yes
 set -g default_user your_normal_user
 set -g theme_color_scheme dark
+set -g theme_right_prompt yes
 ```
 
 **Title options**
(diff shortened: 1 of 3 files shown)


bobthecow:
I feel like a fishier way to do this is to define your own `fish_right_prompt`.

bobthecow:
Closing this issue. I've added documentation to the README about the fishy way to accomplish this :)

https://github.com/oh-my-fish/theme-bobthefish#overrides

rickard-von-essen:
Sorry for my radio silence. I've been struggling to follow advice but my conclusion is that it doesn't work for me.

I'm using [fisherman](http://fisherman.sh/) and it symlinks `~/.config/fish/functions/fish_right_prompt.fish` to the theme plugin, in this case `~/.config/fisherman/bobthefish/fish_right_prompt.fish`. 

I have struggled to find a way to override this haven found how.

PR #60, opened 2016-08-28
Fix #52

### LINKED ISSUE #52: Prompt shows wrong Python version in virtualenv
Hi!

When I create a Python 3 virtualenv using `venv` or `virtualenv` and activate it, the prompt shows the wrong version 2.

I think this is because `__bobthefish_virtualenv_python_version` uses the name of the python executable to determine the version. This is (im my case at least) always `$venvdir/bin/python`, regardless of Python version and the prompt assumes Python 2.

Wouldn't it be possible to use the output of `python --version` instead?

rominf:
It's working with openSUSE 13.2, `pyenv`, `pypy`, Python's 2 and 3 versions (irrespective of an executable name).

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -536,17 +536,12 @@ function __bobthefish_prompt_vi -S -d 'Display vi mode'
 end
 
 function __bobthefish_virtualenv_python_version -S -d 'Get current python version'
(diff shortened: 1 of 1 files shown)


bobthecow:
Thanks!

PR #69, opened 2016-12-23
Add variable to ignore certain folder for hg project search

rominf:
It's set to `$HOME` by default. Also I considered the recommendation about setting some mercurial environmental variables from this letter: https://www.mercurial-scm.org/pipermail/mercurial/2014-November/047687.html

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -757,9 +757,6 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   # Save the last status for later (do this before the `set` calls below)
   set -l last_status $status
 
-  # hg settings
-  set -l theme_hg_ignored_dirs $HOME
-
   # Powerline glyphs
   set -l __bobthefish_branch_glyph            \uE0A0
   set -l __bobthefish_right_black_arrow_glyph \uE0B0
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -24,6 +24,7 @@
 #     set -g theme_display_vagrant yes
 #     set -g theme_display_docker_machine no
 #     set -g theme_display_hg yes
+#     set -g theme_hg_ignored_dirs
 #     set -g theme_display_virtualenv no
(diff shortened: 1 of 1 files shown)


fish_prompt.fish
```diff
@@ -753,6 +757,9 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
   # Save the last status for later (do this before the `set` calls below)
   set -l last_status $status
 
+  # hg settings
+  set -l theme_hg_ignored_dirs $HOME
```
bobthecow: This will override whatever the user sets, right? If we're going to do this, let's change it to…

```
set -q theme_hg_ignored_dirs
  or set -l theme_hg_ignored_dirs $HOME
```

?

rominf:
Thank you for pointing me out. Actually, that's correct. But I left this variable anyway, because some repositories can be so big, that shell becomes unresponsible.
Is it OK now?

fish_prompt.fish
```diff
@@ -119,7 +120,8 @@ end
 function __bobthefish_hg_project_dir -S -d 'Print the current hg project base directory'
   [ "$theme_display_hg" = 'yes' ]; or return
   set -l d $PWD
-  while not [ $d = / ]
+  set stop_on_dirs / $theme_hg_ignored_dirs
+  while not contains $d $stop_on_dirs
```
bobthecow: I don't think we need variable assignment here. Just

```
while not contains $d / $theme_hg_ignored_dirs
  ...
```

should do the trick?

rominf:
Sure

bobthecow:
This should be covered in a more generic way for both git and hg by #111. 

One thing from this PR that might be relevant is this bit:

```
+  set -x HGRCPATH ''  # Disable hg plugins
+  set -x HGPLAIN 1    # Ensure parsable output of hg
```

Can you explain that to me? What does hg do without those? Should we be using those all the time anyway?

PR #72, opened 2017-01-09
Display user in terminal title

kephas:


kephas:
I didn't see how, but avoiding the code duplication would obviously be nice here.

fish_title.fish
```diff
@@ -3,7 +3,19 @@
 #     set -g theme_title_display_path no
 #     set -g theme_title_use_abbreviated_path no
 
+function __bobthefish_user -S -d 'Display actual user if different from $default_user'
+  if [ "$theme_display_user" = 'yes' ]
```
bobthecow: let's make this `$theme_title_display_user` so they can be controlled independently.

fish_title.fish
```diff
@@ -3,7 +3,19 @@
 #     set -g theme_title_display_path no
 #     set -g theme_title_use_abbreviated_path no
 
+function __bobthefish_user -S -d 'Display actual user if different from $default_user'
```
bobthecow: … and then call this `__bobthefish_title_user` :)

bobthecow:
actually, don't worry about it. i'll merge then make that change.

PR #75, opened 2017-01-17
Use dark letters as foreground for solarized

JelteF:
Most of the colors are quite light and dark on light is much better (IMHO) readable than white on light.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -1162,7 +1162,7 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
       set -l cyan    2aa198
       set -l green   859900
 
-      set colorfg $base3
+      set colorfg $base03
 
       set __color_initial_segment_exit     $base2 $red --bold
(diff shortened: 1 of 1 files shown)


bobthecow:
the light foreground looks great on my box, and is more consistent with the theme.

solarized is specifically designed so that any of the four background colors can be used on top of any accent color, or any accent color can be used on top of any background. if you're having trouble with it, is there a possibility that it's a terminal app config or monitor calibration issue?

JelteF:
They are definitely both readable, it's just that (especially on the green background) the dark is much more easy on my eyes, especially when using a small fontsize.
dark vs light

bobthecow:
Thank you for the suggestion. I'll be sticking with the original color scheme, but luckily, bobthefish is designed so you can work around my opinions :)

Just copy paste your solarized dark config variant into your fish config, change all the lines to `set -g` rather than `set`, and add `set -g theme_color_scheme user` :)

PR #76, opened 2017-01-19
Remove space after virtualenv glyph so version is connected to it

JelteF:


bobthecow:
Mind adding a before / after screenshot?

JelteF:
Before
before
After
after

bobthecow:
For me…

before: 
screen shot 2017-01-19 at 10 22 14 am

after:
screen shot 2017-01-19 at 10 22 05 am

:(

JelteF:
Hmm weird, is it considered as a single or double width character for you. 

Also, what terminal and OS do you have?

I'm runnig Arch linux with xfce4-terminal

bobthecow:
I'm using iTerm2 on OS X. There's a setting in iTerm to "Treat ambiguous-width characters as double width", and ticking that makes it act like yours. But it very not recommended to enable this option :)

screen shot 2017-01-19 at 10 47 46 am

Maybe we need a theme config to set for people whose terminals treat these glyphs as double width. Are you seeing extra space around other glyphs as well?

JelteF:
I have a similar option, with a similar warning, and its turned off. However, turning it on doesn't change anything. When copying the two characters to the command line they also both take a single space.
commandline

Looking better at your screenshot it actually seems like your character width is so small that the python glyph takes more than a single space (based on the width of your cursor).

bobthecow:
I think we need to have a setting for whether to include spaces after certain glyphs. Some shells will never look good with the spaces, some shells will never look good without them.

PR #79, opened 2017-01-30
Improves interactive shell performance

### LINKED ISSUE #70: This theme slows down interactive shell
I have been using this theme for quite some time, but a few months ago I noticed that my shell is sluggish when I turn my laptop into power save mode. Any other theme I have tried today doesn't introduce any noticeable slowdown. Backspace operation is the slowest, if I have a 50-character-long line and I press and hold backspace for 3 seconds and release it, the shell will continue deleting characters for 4-5 seconds more, so it can remove the whole line while I wanted to remove just a few words.

System info:

* Arch Linux x64 (Core i7 in power save mode running at 800MHz)
* Fish 2.4.0
* Latest bobthefish

frol:
Partially fixes #70.

It is still noticeably lagging, but not as bad as it used to be.

/cc @ycardon

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -200,19 +200,18 @@ function __bobthefish_start_segment -S -d 'Start a prompt segment'
   set -l fg $argv[1]
   set -e argv[1]
 
-  set_color normal # clear out anything bold or underline...
-  set_color -b $bg
-  set_color $fg $argv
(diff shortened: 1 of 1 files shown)


bobthecow:
Thanks for looking into this! Closing this one in favor of the core fix :)

PR #83, opened 2017-02-27
Display superuser glyph when using fakeroot

LaurentTreguier:
Using `id -u` displays the current active id instead of the actual real one. This means that the superuser glyph will be shown when using the `fakeroot` utility to simulate being root.
Though I don't know if it was an intented feature not to show it in such cases...

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -383,7 +383,7 @@ function __bobthefish_prompt_status -S -a last_status -d 'Display symbols for a 
     and set nonzero $__bobthefish_nonzero_exit_glyph
 
   # if superuser (uid == 0)
-  [ (id -u $USER) -eq 0 ]
+  [ (id -u) -eq 0 ]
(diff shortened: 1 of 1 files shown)


bobthecow:
It's not intended… I don't have a strong opinion whether to show the superuser glyph when using `fakeroot`. Maybe it makes sense to show the glyph, but use a different color to disambiguate?

LaurentTreguier:
It would probably make sense, though I don't know which color should be used... Maybe staying with green for `fakeroot` and red or yellow for real superpowers ?

bobthecow:
I was thinking the other way around. bobthefish uses red for "no write permission" on directories; it'd make sense to me to use the same color for "no real root permissions". If we go that route, we could just reuse `$__color_initial_segment_exit`.

LaurentTreguier:
Yes, that makes more sense indeed if bobthefish uses red for "no root permission".

bobthecow:
:+1:

PR #90, opened 2017-05-03
Moved Git chars to glyph variables

nemoDreamer:
Even if no suitable Powerline/NerdFont alternatives are found, this PR brings the Git implementation closer to the standard of the Vagrant/etc ones, and would be a step towards allowing for external overrides.


bobthecow:
Thank you!

Issue #62, opened 2016-09-03
Add option to move cursor in a new line


diestrin:
Currently in my screen there's almost no space to type. I'd be awesome if I can move the cursor to a new line always.

![-v-h-p-1-w-wu 2016-09-03 13-30-02](https://cloud.githubusercontent.com/assets/909327/18227233/abe3f65e-71da-11e6-9e9d-7845314a475c.png)

bobthecow:
Unfortunately I haven't found a way to do this which doesn't screw with the right prompt or make things really awkward when clearing the screen. I'm happy to accept pull requests if someone can figure it out :)

nicbet:
Here is what works for me: 

I've added a simple echo after line 1326 
```
  # Insert newline
  echo -e ""

  # Start each line with a blank slate
  set -l __bobthefish_current_bg

  __bobthefish_maybe_display_colors
```

<img width="870" alt="screen shot 2017-03-28 at 7 50 25 pm" src="https://cloud.githubusercontent.com/assets/718340/24432388/e163e3c2-13ef-11e7-9c09-5ee708317a7f.png">

bobthecow:
@nicbet Is this what @diestrin was looking for? I read it as wanting the prompt to stay on the line above, but the cursor to move down to the next line.

diestrin:
Yeah, thanks @nicbet for the response, but that's not quite what I was looking for 👍

druchoo:
@diestrin, @bobthecow, I took a try at this with #92. Let me know what you think.

bobthecow:
This feature has been added with #92. Thanks @druchoo!

PR #92, opened 2017-06-11
Feature/cursor on a new line

druchoo:
This will solve #62. I hope it's not too hacky for you. The trick here is the new line needs to start with a character otherwise the right prompt will only show after typing one character. Other than that, I haven't seen any issues.

Here are some pics:

* Default 
```
set -g theme_newline_cursor yes
```
cursor_on_a_newline_righ_arrow_glyph

* No right arrow glyph on new line
```
set -g theme_newline_cursor yes
set -g theme_newline_cursor_arrow_glyph no
```
cursor_on_a_newline

* No right arrow glyph on top line
```
set -g theme_newline_cursor yes
set -g theme_newline_cursor_top_arrow_glyph no
```
cursor_on_a_newline_righ_arrow_glyph_no_top_right_arrow_glyph

* Minimalist
```
set -g theme_newline_cursor yes
set -g theme_newline_cursor_arrow_glyph no
set -g theme_newline_cursor_top_arrow_glyph no
```
cursor_on_a_newline_no_right_arrow_glyphs

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,6 @@ This theme is based loosely on [agnoster][agnoster].
      * Background jobs (**`%`**)
      * You currently have superpowers (**`$`**)
      * Cursor on newline
-       * Prefix newline with right arrow glyph
-       * Keep right arrow glyph on live above newline
  * Current vi mode
  * `User@Host` (unless you're the default user)
  * Current RVM, rbenv or chruby (Ruby) version
@@ -96,8 +94,6 @@ set -g theme_color_scheme dark
 set -g fish_prompt_pwd_dir_length 0
 set -g theme_project_dir_length 1
 set -g theme_newline_cursor yes
-set -g theme_newline_cursor_arrow_glyph no
-set -g theme_newline_cursor_top_arrow_glyph no
 ```
 
 **Title options**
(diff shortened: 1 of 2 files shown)


diestrin:
I like it. But I want to challenge you even further :p Is it possible to keep the time in the first line?
Anyway, thanks for this PR

druchoo:
I'll see what I can do. https://github.com/fish-shell/fish-shell/issues/1706 looks promising but not sure how much modification will be required.

bobthecow:
This implementation, unfortunately, has the same issues as the attempt I commented on in #62. Specifically:

1. When you clear the buffer (cmd+k in iterm or terminal app, not `clear` as far as i can tell) you lose the first line of your left prompt.
2. The right prompt is on the wrong line and sometimes kind of wonky.

The hacks in that fish-shell issue probably aren't worth doing, unless you're also going to reproduce the right prompt hiding logic (if the left prompt or text you're typing gets too long and encroaches on the right prompt, fish automatically hides the entire right prompt). As it stands, it looks like their code just makes a mess of things. Additionally, their hacks will just exacerbate issues with ambiguous unicode character widths :(

At this point it doesn't look like fish is adding multi-line right prompt support any time soon, so I'm probably okay implementing it as an option for people, even with the two issues above.

druchoo:
Hey @bobthecow,

Thanks for reviewing.

1.  I figured you meant `clear` command when you referenced `clearing the screen`. Honestly, I didn't know about `CMD+K` so i'll have to looking into that issue.
2. Didn't realize the requirement of having the right prompt on first line. Figured it should be on second in case the first line became too long. At least this is how I wanted it. What other wonky'ness are you seeing?

bobthecow:
"in case the line is too long" is handled by the default auto-hide logic for the right prompt :)

In keeping with bobthefish's goal to be _opinionated_ but _flexible_, let's do this with one option, not three. I'm leaning toward:

```fish
set -g theme_cursor_on_new_line yes
```

Then:
* Always show the arrow at the end of the first line
* Always show a minimalist arrow at the start of the next line

Something like this:

```fish
function __bobthefish_finish_segments -S -d 'Close open prompt segments'
  if [ "$__bobthefish_current_bg" != '' ]
    set_color normal
    set_color $__bobthefish_current_bg
    echo -ns $__bobthefish_right_black_arrow_glyph ' '
  end

  if [ "$theme_newline_cursor" = 'yes' ]
    echo -ens "\n"
    set_color $fish_color_autosuggestion
    if [ "$theme_powerline_fonts" = "no" ]
      echo -ns '> '
    else
      echo -ns "$__bobthefish_right_arrow_glyph "
    end
  end

  set_color normal
  set __bobthefish_current_bg
end
```

druchoo:
@bobthecow, This looks good to me. I actually like `right_arrow_glyph` better than `right_black_arrow_glyph` :-) It's very subtle

cursor_on_a_newline_bobthecow


New commit to reflect your changes.

bobthecow:
Thanks for your contribution!

Issue #86, opened 2017-03-22
Add ability to customize glyphs


nemoDreamer:
It would be awesome if one could override the glyphs, similar to the `$theme_` options.

I don't know the correct "`fish`" for this, but something like "set locally from global or default":

`set -l __bobthefish_my_var (($theme_my_var) -o \u2191)` ?

or `if test ! -n $__bobthefish_my_var; set -l __bobthefish_my_var \u2191`, but that would allow for overriding w/ `''`...

nemoDreamer:
Currently, they're being set w/ a forced local scope, which means you can't pre-define globally:

```fish
# Powerline glyphs
set -l __bobthefish_branch_glyph            \uE0A0
set -l __bobthefish_right_black_arrow_glyph \uE0B0
set -l __bobthefish_right_arrow_glyph       \uE0B1
set -l __bobthefish_left_black_arrow_glyph  \uE0B2
set -l __bobthefish_left_arrow_glyph        \uE0B3

# Additional glyphs
set -l __bobthefish_detached_glyph          \u27A6
set -l __bobthefish_tag_glyph               \u2302
set -l __bobthefish_nonzero_exit_glyph      '! '
set -l __bobthefish_superuser_glyph         '$ '
set -l __bobthefish_bg_job_glyph            '% '
set -l __bobthefish_hg_glyph                \u263F

# Python glyphs
set -l __bobthefish_superscript_glyph       \u00B9 \u00B2 \u00B3
set -l __bobthefish_virtualenv_glyph        \u25F0
set -l __bobthefish_pypy_glyph              \u1D56

# Ruby glyps
set -l __bobthefish_ruby_glyph              ''

# Vagrant glyphs
set -l __bobthefish_vagrant_running_glyph   \u2191 # ↑ 'running'
set -l __bobthefish_vagrant_poweroff_glyph  \u2193 # ↓ 'poweroff'
set -l __bobthefish_vagrant_aborted_glyph   \u2715 # ✕ 'aborted'
set -l __bobthefish_vagrant_saved_glyph     \u21E1 # ⇡ 'saved'
set -l __bobthefish_vagrant_stopping_glyph  \u21E3 # ⇣ 'stopping'
set -l __bobthefish_vagrant_unknown_glyph   '!'    # strange cases
```

nemoDreamer:
[...]

PR #99, opened 2017-08-02
Git glyphs for Nerd Fonts

nemoDreamer:
Ok, @bobthecow ,
I've used these for a long time now, and I think they address your concerns from #86 
Wanna take this branch for a spin?


bobthecow:
:+1: i've used some of these since #86 :)

I'm still not sold on the heavy question mark for untracked. `save` for staged hasn't felt quite right, but I don't have any suggestions for making it better.

nemoDreamer:
❤️

EvanCarroll:
@bobthecow is the question mark for untracked even documented anywhere? Found this googling for it.

PR #101, opened 2017-08-08
Better hybrid mode support for theme_display_vi

cmrn:
Currently the only way to enable the vi mode indicator is to be using the vanilla `fish_vi_key_bindings` or the (undocumented) `fish_hybrid_key_bindings`.

This PR enables the vi mode indicator by default for people using the `hybrid_bindings` function from the [fish documentation](https://fishshell.com/docs/current/index.html#editor). It also adds an option to either enable and disable the indicator no matter what key bindings function is in use.

This is useful for people who have either copied the function from the fish documentation, or have a custom key bindings function.


bobthecow:
Looks good. Thank you!

PR #102, opened 2017-08-08
Added support for showing current Kubernetes context

uberstudio:
This commit adds support for reading the user's current Kubernetes context used by `kubectl` and displaying it in the command prompt.

This is extremely useful for people managing multiple Kubernetes clusters from the same machine.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -23,6 +23,7 @@
 #     set -g theme_git_worktree_support yes
 #     set -g theme_display_vagrant yes
 #     set -g theme_display_docker_machine no
+#     set -g theme_display_k8s_context no
 #     set -g theme_display_hg yes
 #     set -g theme_display_virtualenv no
 #     set -g theme_display_ruby no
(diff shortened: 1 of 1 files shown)


cristian04:
+1

todaywasawesome:
The `KUBECONFIG` environmental variable is reliable if you've switched contexts mid-session. Would it be too laborious to check if that is empty and if it is grep the default config? That would be reliable, fast, and avoid invoking kubectl. 

Or maybe when fish initializes we could have it load the KUBECONFIG environmental variable? Is that an option. 

See my experiments:

` !  ~/G/demochat  ➦ 90f4447d  grep current-context $KUBECONFIG  Fri Dec 22 21:41:00 2017
current-context: mycluster
 ~/G/demochat  ➦ 90f4447d  gcloud container clusters get-credentials meetup --zone us-west1-b --project firstkubernetes-176201
Fetching cluster endpoint and auth data.
kubeconfig entry generated for meetup.
 ~/G/demochat  ➦ 90f4447d  kubectl config current-context
gke_firstkubernetes-176201_us-west1-b_meetup
 ~/G/demochat  ➦ 90f4447d  grep current-context $KUBECONFIG
current-context: gke_firstkubernetes-176201_us-west1-b_meetup
 ~/G/demochat  ➦ 90f4447d  `

bobthecow:
Okay, check out [the k8s branch](https://github.com/oh-my-fish/theme-bobthefish/tree/k8s)!

This _should_ be working as expected. Please let me know if it isn't, because I don't actually have kubernetes set up on this machine so I faked it all for testing :P 

That implementation (based on this PR, with some bonus goodness) adds support for any number of `:`-separated paths in `$KUBECONFIG`, plus `~/.kube/config`. It supports fallback, i.e. if one of the more specific configs doesn't have a context, it checks the more general ones as well. From reading the `kubectl` source, this _seems_ to be how that works as well.

todaywasawesome:
Tested and working great! The [k8s branch](https://github.com/oh-my-fish/theme-bobthefish/tree/k8s) fast and works well. 

screen shot 2017-12-27 at 2 03 15 pm

bobthecow:
Kubernetes context support has shipped!

Let me know (in a new issue or pull request, SVP) of any improvements you'd like to see.

PR #111, opened 2017-10-16
Add support for ignoring VCS directories

jderehag:
Related to PR #69 but works for both vcs backends (git and hg) and also supports globbed paths.

Signed-off-by: Jesper Derehag <jderehag@hotmail.com>

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -92,8 +92,19 @@ function __bobthefish_pretty_parent -S -a current_dir -d 'Print a parent directo
   string replace -ar '(\.?[^/]{'"$fish_prompt_pwd_dir_length"'})[^/]*/' '$1/' "$parent_dir/"
 end
 
+function __ignore_vcs_dir -d 'Identifies if dir is to be ignored for VCS highlighting'
+  for path_ in $theme_vcs_ignore_paths
+    if [ (string match (realpath $path_ ^/dev/null) (realpath $PWD ^/dev/null)) ]
+      echo 1
+      break
+    end
+  end
+end
+
 function __bobthefish_git_project_dir -S -d 'Print the current git project base directory'
   [ "$theme_display_git" = 'no' ]; and return
+  [ (__ignore_vcs_dir) ]; and return
+
   if [ "$theme_git_worktree_support" != 'yes' ]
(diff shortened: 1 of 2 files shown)


fish_prompt.fish
```diff
@@ -92,8 +92,19 @@ function __bobthefish_pretty_parent -S -a current_dir -d 'Print a parent directo
   string replace -ar '(\.?[^/]{'"$fish_prompt_pwd_dir_length"'})[^/]*/' '$1/' "$parent_dir/"
 end
 
+function __ignore_vcs_dir -d 'Identifies if dir is to be ignored for VCS highlighting'
+  for path_ in $vcs_ignore_paths
+    if [ (string match -r (realpath $path_ ^/dev/null) (realpath $PWD ^/dev/null)) ]
```
bobthecow: this probably should be a regular match (i.e. glob) rather than a regex match.

fish_prompt.fish
```diff
@@ -92,8 +92,19 @@ function __bobthefish_pretty_parent -S -a current_dir -d 'Print a parent directo
   string replace -ar '(\.?[^/]{'"$fish_prompt_pwd_dir_length"'})[^/]*/' '$1/' "$parent_dir/"
 end
 
+function __ignore_vcs_dir -d 'Identifies if dir is to be ignored for VCS highlighting'
+  for path_ in $vcs_ignore_paths
```
bobthecow: we try to use a `theme_` prefix for as many theme related config options as possible. should this use that prefix as well?

fish_prompt.fish
jderehag: doh, of course you are right... I'll fix it..

fish_prompt.fish
jderehag: I suspected as much but wasnt sure, I will rename it..

jderehag:
@bobthecow Ok, updated according to comments and also added some notes in the readme.

bobthecow:
Thanks, merged!

After playing with this for a bit, I decided it needed an implicit `/*` on the end of every pattern. Otherwise, the repo root would not have a vcs segment in the prompt, but subdirectories inside the repo would show it, and that's super weird :)

PR #112, opened 2017-10-17
show kubernetes context

tanner-bruce:


tanner-bruce:
yeah you are probably right, I get 0 performance hit though. Get a new PC if you do :)

uberstudio:
This code is for a command prompt theme.
There really is no reason for it take any performance hits it can avoid.

I run terminal emulators on many systems I manage, many of which are low-end or embedded systems.
Using your code, I might experience a 1-2 second hang before each new line comes up.

tanner-bruce:
@uberstudio 

Your way will not take into account the KUBECONFIG environment variable. It may be faster to grep the ~/.kube/config file, but it is incorrect. Although I see your point, I doubt you are running kubectl on embedded systems.

bobthecow:
as a general rule, bobthefish tries to be fast above all else. for a great example of this, check out the craziness in `__bobthefish_hg_project_dir` just to avoid checking whether a directory is an `hg` repo :)

additionally, things that are slow are defaulted to off and must be opted into (e.g. `theme_git_worktree_support yes`, `theme_display_vagrant yes`, `theme_display_hg yes`, etc). if y'all want this enabled by default, we probably have to be able to do it using only builtins.

so, is there a way to support `KUBECONFIG` without using `kubectl`?

todaywasawesome:
From my tests running `kubectl config current-context` is slow the first time and fast subsequent times. I don't know of a way to bypass kubectl to get this information but looking at the code might lead to an answer. 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/config/current_context.go

This might be the first time I've actually seen go, so I'm afraid I can't help more than that :(

bobthecow:
As far as I can tell from reading the source, there are three ways to configure kubernetes (as long as we ignore a couple of deprecated paths, which I’m happy to do): `~/.kube/config`, any number of `:`-separated paths in `$KUBECONFIG` environment variable, or passing in `--kubeconfig=foo` to `kubectl`. 

We can ignore the last way, because obviously nobody is using that here. 

The other two ways are almost trivial to implement ourselves :)

bobthecow:
Okay. Closing this one in favor of #102 because I think this is totally possible. Continue conversation over there!

Issue #114, opened 2017-10-25
Feature Req: Please support asdf


jschank:
it would be nice to show supported package management information, from asdf.
asdf, is a more generic tool that does what nvm, rbenv, rvm, virtualenv all do.

bobthecow:
I don’t use asdf, but pull requests are welcome :)

jschank:
Hmm, i just started using it myself. I’ll see what i can figure out. Could be an interesting problem.
Thanks,
Schank

jschank:
HI there. I am working on adding the asdf functionality. I have forked the theme, and I have it installed locally. But I am unfamiliar with the best practices for contributing... For example, Should I create a branch on my forked repo, for the new work, Or should the commits just go straight onto master?

jschank:
Heh, I think I figured it out. I have created a PR to add in ruby from asdf to the ruby prompt.
Note: I notice that by default, the theme attempts to not report the ruby version when it is the 'system' version. I think I would prefer to be able to always see what version of ruby is currently active (even if it is the global/system version - because the global version can be changed as well) Perhaps, the configuration setting for `set -g theme_display_ruby no` should allow options of `yes|no|always`

PR #119, opened 2017-12-21
#114 Add asdf ruby version to ruby version prompt

jschank:
Followed established pattern for chruby and rbenv, adding the ruby
version number from asdf to the ruby prompt, only when the local
ruby version differs from the system ruby.

Note that asdf tracks its versions in a dot file (.tool-versions) so
the concept of what is global/system vs what is local simply depends
upon what file asdf is using to determine the correct version.
Fortunately, asdf reports the .tool-versions file when querying for
the current version.

The downside to this is that correct behavior of this patch will
depend upon the format asdf uses for their provenance string.

See https://github.com/asdf-vm/asdf for details about asdf

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -586,6 +586,15 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
     [ "$ruby_version" = "$global_ruby_version" ]; and return
   else if type -q chruby
     set ruby_version $RUBY_VERSION
+  else if type -q asdf
+    set -l asdf_ruby_version_str (asdf current ruby ^ /dev/null)
+    if test $status -eq 0
+      set -l asdf_system_origin "(set by $HOME/.tool-versions)"
+      set -l asdf_current_version_provenance (string split -m 1 ' ' $asdf_ruby_version_str)[2]
+      set ruby_version (string split -m 1 ' ' $asdf_ruby_version_str)[1]
+
+      [ "$asdf_current_version_provenance" = "$asdf_system_origin" ]; and return
+    end
   end
   [ -z "$ruby_version" ]; and return
   __bobthefish_start_segment $__color_rvm


bobthecow:
Thanks!

bobthecow:
I reworked the implementation a bit for performance, and added a comment about the asdf provenance format caveat. Mind giving it a shot and confirming that it all works as expected?

jschank:
Hmm, I get an exception thrown, when I use bobthefish master branch.

Lemme read about read…. :)

looks like this.


read: Variable name '2.4.2 (set by /Users/jschank/.tool-versions)' is not valid. See `help identifiers`.
~/.config/fish/functions/fish_prompt.fish (line 608):
    read -l asdf_ruby_version asdf_provenance -- $asdf_current_ruby
    ^
in function '__bobthefish_prompt_rubies'
called on standard input

in function 'iterm2_fish_prompt'
called on line 61 of file ~/.iterm2_shell_integration.fish

in command substitution
called on line 59 of file ~/.iterm2_shell_integration.fish

in command substitution
called on line 59 of file ~/.iterm2_shell_integration.fish

in function 'fish_prompt'
called on standard input

in command substitution
called on standard input


       readread - read line of input into variables
        -

   Synopsis
       read [OPTIONS] [VARIABLES...]

read: Type 'help read' for related documentation

jschank:
I think I see.
`$ asdf current ruby`
will yield this:
`2.4.2 (set by /Users/jschank/.tool-versions)`

I think read is attempting to create a variable called '2.4.2', which seems wrong.

bobthecow:
oh, yep. totally my fault. i tested with a fake (`foo bar`) which didn't catch this. lemme fix really quick :)

bobthecow:
Ok. Should be fixed by d8d7a544. And that one's even more concise!

jschank:
Just pulled and tested, Seems to be working fine!

bobthecow:
Thanks for checking it out. And thanks for your help with this :)

Issue #97, opened 2017-07-17
Add NVM support


Yuri-M-Dias:
First of all, thanks for the amazing work on this theme!

Secondly, I'd like to request a Node Version Manager integration. I've noticed that there's only one theme that has this currently, and it'd be pretty helpful. 
I can add it myself over the next weekend if nobody wants to take it. I can add it like the ruby/python managers and make it optional with an env var.

bobthecow:
Hi! Sorry I missed this one. Yes, NVM support would be great :)

rcalabro:
+1

PR #128, opened 2018-01-28
Ability to show current Node.js version

imposibrus:
Possible fix for #97


bobthecow:
In all the other "version" segments (virtualenv, rvm, etc) the segment isn't showing the current version of the language, it's showing the current state of the virtual environment. That this _happens_ to show the language version is a bit irrelevant. 

For this to be consistent with those, it would probably show the version supplied by NVM (or other node environment manager), along with any equivalent to ruby bundles.

Additionally, if the current environment version is the "system" or "default", the other language segments don't show at all. This is in keeping with the bobthefish principle that default states should be omitted whenever possible, so that emphasis is placed on what is _different_, not what is always the same.

So to move forward with this… What is setting your node version?

rcalabro:
I think the best option would be to use the active NVM version.

PR #131, opened 2018-02-15
Make it possible to show basename in prompt

grimpy:


README.md
```diff
@@ -114,7 +114,7 @@ set -g theme_newline_cursor yes
 - `theme_display_k8s_context`. By default the current kubernetes context is shown (`> kubectl config current-context`). Use `no` to hide the context.
 - `theme_show_exit_status`. Set this option to yes to have the prompt show the last exit code if it was non_zero instead of just the exclamation mark.
 - `theme_git_worktree_support`. If you do any git worktree shenanigans, setting this to `yes` will fix incorrect project-relative path display. If you don't do any git worktree shenanigans, leave it disabled. It's faster this way :)
-- `fish_prompt_pwd_dir_length`. bobthefish respects the Fish `$fish_prompt_pwd_dir_length` setting to abbreviate the prompt path. Set to `0` to show the full path, `1` (default) to show only the first character of each parent directory name, or any other number to show up to that many characters.
+- `fish_prompt_pwd_dir_length`. bobthefish respects the Fish `$fish_prompt_pwd_dir_length` setting to abbreviate the prompt path. Set to `-1` to show basename, `0` to show the full path, `1` (default) to show only the first character of each parent directory name, or any other number to show up to that many characters.
```
bobthecow: in this case, we're using a builtin fish config var to mean the exact same thing in bobthefish as it does in fish's default prompt. overriding it to add extra meaning is a bit weird. instead, we can add a separate config var that is theme-specific (i.e. has a `theme_` prefix) and check that.

grimpy:
@bobthecow I've updated the PR to address your comment.
Personally codewise I like this less altough I see your point. If you decide to merge it feel free to squash it I made it 2 commits so the 2 implementation can be compared easily

grimpy:
@bobthecow anything else you want me to do in this?

PR #133, opened 2018-02-17
Add verbose option for Git dirty state

PhilipF5:
This adds the one piece of information I found missing from the Git display: an indicator of how many lines have been added/removed when the repo is in a dirty state. Liquidprompt for Bash does this, and I find it very helpful information to see at a glance. I implemented it as a configurable option similarly to `$theme_display_git_ahead_verbose` and defaulted it to off. I also verified that if the verbose options are turned on for both "ahead" and "dirty," it doesn't screw up the display.

Credit to [hughes at Stack Overflow](https://stackoverflow.com/questions/9933325/is-there-a-way-of-having-git-show-lines-added-lines-changed-and-lines-removed#comment55779320_9933440) for the Git and awk commands I used to retrieve the line change stats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ set -g theme_display_git no
 set -g theme_display_git_dirty no
 set -g theme_display_git_untracked no
 set -g theme_display_git_ahead_verbose yes
+set -g theme_display_git_dirty_verbose yes
 set -g theme_git_worktree_support yes
 set -g theme_display_vagrant yes
 set -g theme_display_docker_machine no
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -21,6 +21,7 @@
 #     set -g theme_display_git_dirty no
 #     set -g theme_display_git_untracked no
 #     set -g theme_display_git_ahead_verbose yes
+#     set -g theme_display_git_dirty_verbose yes
 #     set -g theme_git_worktree_support yes
 #     set -g theme_display_vagrant yes
 #     set -g theme_display_docker_machine no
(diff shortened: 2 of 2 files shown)


bobthecow:
:+1:

It seems like it would be a bit cleaner to do something like this?

```diff
diff --git a/fish_prompt.fish b/fish_prompt.fish
index 98301f3..b483165 100644
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -717,6 +717,9 @@ function __bobthefish_prompt_git -S -a current_dir -d 'Display the actual git st
     set -l show_dirty (command git config --bool bash.showDirtyState ^/dev/null)
     if [ "$show_dirty" != 'false' ]
       set dirty (command git diff --no-ext-diff --quiet --exit-code ^/dev/null; or echo -n "$__bobthefish_git_dirty_glyph")
+      if [ "$dirty" -a "$theme_display_git_dirty_verbose" = 'yes' ]
+        set dirty "$dirty"(__bobthefish_git_dirty_verbose)
+      end
     end
   end
```

(I didn't actually run this, so I'm not sure it works, but it seems right 😛)

PhilipF5:
You're right, that does work and make it simpler. I tried something similar the first go around, but couldn't get it to work, probably because I was missing something. Added a second commit with this simplification.

bobthecow:
Thanks!

bobthecow:
To match the spirit of bobthefish a bit closer, I've removed the `+0` and `-0` states, so it only ever shows interesting (non-zero) information :)

PR #135, opened 2018-02-23
Fix issue with filtering out +0/-0

PhilipF5:
I discovered that there are some situations where a dirty state can be reported by `git_dirty_verbose` as `+0/-0`. At minimum, I have observed this happening in merge conflicts, and when there is modified, uncommitted content in a submodule. Line 247 would remove the `+0/` but leave behind the `-0` because then it wouldn't match the `/-0` in the second string replacement. I added a check specifically to handle the `+0/-0` special case.


bobthecow:
Thanks! I've pushed a fix for this that uses a regex rather than piping through `string replace` three times :)

Issue #129, opened 2018-01-31
theme_display_user does not honor default_user


letheed:
Just ran `omf update` and my username now appears in the prompt.

I believe [this](https://github.com/oh-my-fish/theme-bobthefish/blob/master/fish_prompt.fish#L525) should be the culprit. It should probably be something like:

```fish
[ \( "$theme_display_user" = 'yes' -a "$USER" != "$default_user" \) -o -n "$SSH_CLIENT" ]
```

bobthecow:
I think the logic was wrong before :)

Your user will now show if _any_ of the following are true:

- you set `theme_display_user` to "yes" (i.e. opted in to "show my username")
- you set a `default_user`, and the current user is not that user
- you're SSHd into another machine

If you want the old behavior back, you can stop setting `theme_display_user` and it'll go back to the old way.

You _could_ argue that there's no longer a way to opt out of the username segment entirely, and you'd be right. But it doesn't sound like that's what you're looking for?

letheed:
Ah, so `default_user` implies `theme_display_user` now... I see, well that's a bit counter intuitive, I think. I found the old behavior more logical.

Regardless, thank you for pointing it out. And thank you for your theme!

schaten:
> there's no longer a way to opt out of the username segment entirely

It would be great if you could either change this or add a hint in the readme that `theme_display_user` and `theme_display_hostname` don't apply for ssh-sessions.

I would propose adding a variable `theme_display_ssh`. The effect would be that, if:
- `"$theme_display_user" = 'yes'`: Always show username, independent of `theme_display_ssh`
- `"$theme_display_user" = 'no'` and `"$theme_display_ssh" = 'yes'`: Show username only when ssh-session is detected.
- `"$theme_display_user" = 'no'` and `"$theme_display_ssh" = 'no'`: Never show username.
The same applies for the hostname.

Pull Request #139. :)

bobthecow:
Closing as it's now possible (and more obvious via the docs!) to configure all permutations thanks to #139.

Thanks @schaten!

PR #139, opened 2018-03-18
Hide user@hostname when using SSH

schaten:
Setting `theme_display_user` and `theme_display_hostname` has no effect when one is logged in via SSH.

I added a switch `theme_display_ssh` that does the following: If
- `"$theme_display_ssh" = 'yes'`: Show user- and hostname when either an ssh-session is detected or `theme_display_user` and `theme_display_hostname` are set to yes.
- `"$theme_display_ssh" = 'no'`: Don't show user- and hostname if showing those isn't enabled.

`theme_display_ssh` defaults to 'no'. This has the advantage of preventing confusion when one has set `theme_display_user` to no explicitly and is logged in via ssh. ;)

Related to #129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ set -g theme_display_virtualenv no
 set -g theme_display_ruby no
 set -g theme_display_user yes
 set -g theme_display_hostname yes
+set -h theme_display_ssh yes
 set -g theme_display_vi no
 set -g theme_display_date no
 set -g theme_display_cmd_duration yes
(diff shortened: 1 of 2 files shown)


bobthecow:
if no config overrides have been set, we want to show username and host when using SSH. so i'm pretty sure the logic for defaults on `$theme_display_ssh` here is backwards.

* username should be displayed in SSH if `"$theme_display_ssh" != no` and `"$theme_display_user" != no`.
* hostname should be displayed in SSH if `"$theme_display_ssh" != no` and `"$theme_display_hostname" != no`.

does that make sense?

schaten:
I agree concerning the defaults.
> does that make sense?

There are other places (fish_prompt.fish, l.538) where `"$theme_display_user" = yes` is checked. This would mean that `theme_display_user` would have _three_ different effects, depending on whether it ist set to yes, no or something else/not defined. Correct me if I'm wrong.

But I think that is a good idea, and it would also erase the need for `theme_display_ssh`!
I suggest to 'allow' the following three values for theme_display_user and theme_display_host: 'no', 'ssh' and 'yes'.

Logic:
SSH-Session:
if `"$theme_display_user" != no` or `"$theme_display_user" = yes` : show username

Local Session:

if `"$theme_display_user" = yes` : show username.

Same for hostname.

schaten:
Pull request updated.

The defaults should reflect the current behaviour..

bobthecow:
:+1: thank you!

PR #140, opened 2018-03-20
fish_prompt: fix __bobthefish_rvm_info for both rvm installation types

acavalin:
I have several machines with different rvm installations (single-user and multi-user), the theme failed to detect rvm by only considering the `~/.rvm` folder.
I patched the `__bobthefish_rvm_info` function so it considers all possible folders.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -584,12 +584,11 @@ end
 
 function __bobthefish_rvm_info -S -d 'Current Ruby information from RVM'
   # look for rvm install path
-  [ -z "$__rvm_install_path" -a -n "$rvm_path"             ] ; and set -Ux __rvm_install_path "$rvm_path"
-  [ -z "$__rvm_install_path" -a -d /usr/local/rvm          ] ; and set -Ux __rvm_install_path /usr/local/rvm
-  [ -z "$__rvm_install_path" -o ! -d "$__rvm_install_path" ] ; and set -Ux __rvm_install_path ~/.rvm
+  set -q rvm_path
+    or set -l rvm_path ~/.rvm /usr/local/rvm
   
   # More `sed`/`grep`/`cut` magic...
-  set -l __rvm_default_ruby (grep GEM_HOME "$__rvm_install_path"/environments/default | sed -e"s/'//g" | sed -e's/.*\///')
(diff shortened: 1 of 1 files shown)


fish_prompt.fish
```diff
@@ -583,8 +583,13 @@ function __bobthefish_rvm_parse_ruby -S -a ruby_string scope -d 'Parse RVM Ruby
 end
 
 function __bobthefish_rvm_info -S -d 'Current Ruby information from RVM'
+  # look for rvm install path
+  [ -z "$__rvm_install_path" -a -n "$rvm_path"             ] ; and set -Ux __rvm_install_path "$rvm_path"
+  [ -z "$__rvm_install_path" -a -d /usr/local/rvm          ] ; and set -Ux __rvm_install_path /usr/local/rvm
+  [ -z "$__rvm_install_path" -o ! -d "$__rvm_install_path" ] ; and set -Ux __rvm_install_path ~/.rvm
```
bobthecow: these definitely don't need to be `set -Ux`, `set -l` should be sufficient.

it also seems like `~/.rvm` would be higher precedence than `/usr/local/rvm`, but i don't actually use rvm so i don't know how they choose to prioritize multiple configs :) 

does this work for you?

```fish
set -q rvm_path
  or set -l rvm_path /usr/local/rvm ~/.rvm

set -l __rvm_default_ruby (grep GEM_HOME $rvm_path/environments/default ^/dev/null | head -1 | sed -e"s/'//g" | sed -e's/.*\///')
```

acavalin: Yes it works, your solution is cleaner than mine and using `set -l` is more bulletproof in case the user changes the installation type later.

The two installation types are mutually exclusive and the double install is discouraged, but some fool may try to use both at the same time... so `~/.rvm` should take the precedence:

```shell
set -q rvm_path
  or set -l rvm_path ~/.rvm /usr/local/rvm

set -l __rvm_default_ruby (grep GEM_HOME $rvm_path/environments/default ^/dev/null | head -1 | sed -e"s/'//g" | sed -e's/.*\///')
```

acavalin:
I forgot the `^/dev/null`... I'll add it when I return home from vacation

bobthecow:
No worries, I'll take care of it!

PR #150, opened 2018-04-18
added support for conda environments.

gregorgebhardt:


bobthecow:
Conda is an alternative to virtualenv, right? Is there ever a reason to use them together?

For the rubies, bobthefish has one segment that supports several different version mangers and bundlers, since they're generally mutually exclusive. Would it make sense to do the same with pythons?

gregorgebhardt:
I think, you're right. At least, I've never used them together and don't know why this should be necessary. I removed the conda-segment and added the functionality to the virtualenv function.

bobthecow:
Thank you!

PR #163, opened 2018-07-08
Change function bobthefish_display_colors to print results using defined environment values

ohgree:
It's not much of an issue, but while i was customising the glyphs to my liking, i found that `bobthefish_display_colors`'s behavior does not change when i made a difference in the glyphs at `__bobthefish_glyphs.fish`, so i came up with a solution.

I have no idea why exit, su, and jobs glyphs are the only one printed using a 'fixed value', since every other glyphs in this file is represented in variables.


bobthecow:
It hard coded them because glyphs aren’t really meant to be an extension point :P

Issue #164, opened 2018-07-17
customize right_arrow_glyph?


waynehoover:
Hello I would like to customize the $right_arrow_glyph especially when `theme_newline_cursor yes` If I submitted a PR for this would it be accepted?

Or is there a way to do this already? I tried to `set right_arrow_glyph  '>>>'` but this doesn't seem to work.

forkbomb9:
It would be a great improvement

fuzebox1:
You can change the glyphs in .local/share/omf/themes/bobthefish/functions/__bobthefish_glyphs.fish
I would like to see this as a setting as well tough.
Also option to start left prompt with a glyph, and not just the background.

PR #168, opened 2018-09-25
allow custom prompt in new line

tflori:
With `set -g theme_newline_prompt '>>> '` you can define your own prompt when `theme_newline_cursor` is set to `yes`

this solves #164 

Also it is possible to use custom utf-8 characters from nerd font as suggested from @forkbomb9


kitsirota:
Thank you for this PR @tflori, this is a great idea (I was cramming this all on one line before).  I just patched my local fish_prompt.fish out with some minor tweaks to your code to display env variables on the new line.

bobthecow:
Thanks, this has been merged.

PR #169, opened 2018-09-29
Don't show bg_job_glyph after cd command when using autojump

rominf:
`autojump` is executed every time user uses `cd` command. `bobthefish` using `jobs` notices that `autojump` is running and shows `bg_job_glyph` (`%` by default). This commit filters running jobs and shows `bg_job_glyph` only when necessary.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -361,7 +361,7 @@ function __bobthefish_prompt_status -S -a last_status -d 'Display flags for a no
     and set superuser 1
 
   # Jobs display
-  jobs -p >/dev/null
+  jobs -c | grep -v Command | grep -v autojump >/dev/null
     and set bg_jobs 1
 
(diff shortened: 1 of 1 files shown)


bobthecow:
This is pretty expensive to do on every prompt for something that most people won't ever need. Is there a better way to make sure we don't know autojump? If not, we probably need this to be opt-in.

bobthecow:
Yeah, that does sound helpful. 

Is autojump always in the background, or just after you `cd`?

Pitasi:
just after a `cd`

pressing <Enter> immediately after `cd` makes the % disappear

bobthecow:
@Pitasi @rominf I just pushed a change that should exclude autojump without much of a performance hit for anyone, and especially not for people who don't use autojump at all :)

Try it out and let me know if you have any issues!

rominf:
@bobthecow It works nice! Thank you!

Issue #148, opened 2018-04-11
Emacs tramp doesn't recognise the shell prompt


binshengliu:
If the remote server use fish shell and bobthefish theme, emacs would hang when opening a file through ssh because it doesn't recognise the prompt.

Just found a solution [here](https://www.gnu.org/software/emacs/manual/html_node/tramp/Frequently-Asked-Questions.html) in the `tramp does not connect to the remote host` section. Perhaps you want to incorporate this in `fish_prompt`.

This is what I found works.

``` shell
function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
  if test $TERM = "dumb"
    echo "\$ "
    return 0
  end
  # other stuff
end
```

bobthecow:
This is probably something best solved at your end. Adding the following to the end of your `config.fish` should do the trick:

```fish
if test "$TERM" = "dumb"
  function fish_prompt
    echo "\$ "
  end

  function fish_right_prompt; end
  function fish_greeting; end
  function fish_title; end
end
```

binshengliu:
True. Thanks!

federkamm:
For me this only partly solved the problem.  I got still strange behaviour and lagging with TRAMP.  I'm a little bit harsher and use in my config.fish:

    if test "$TERM" = "dumb"
        exec sh
    end

Now, everything runs smoothly again.

azzamsa:
@federkamm solution fixes Trump issue. But it make all my function that contains `start-process` fails.

So I go with @bobthecow solutions.

aegithalos:
This breaks pdf export in org mode, since pdflatex is called like
` fish -c "pdflatex ..." `from emacs.

 I had a hard time to find this out :)

PR #171, opened 2018-10-08
Fix: Use a simple prompt on dumb terminals.

Dietr1ch:
Dumb terminals don't like this fancy prompt.

This is a really a workaround for loading the theme on demand, but it greatly
simplifies managing plugins and allows just relying on current plugin managers.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -948,6 +948,12 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
     # Save the last status for later (do this before the `set` calls below)
     set -l last_status $status
 
+    # Use a simple prompt on dumb terminals.
+    if test $TERM = "dumb"
+        echo "> "
+        return
+    end
+
     __bobthefish_glyphs
     __bobthefish_colors $theme_color_scheme
 


bobthecow:
See #148!

Dietr1ch:
Oh, you also seem to consider this a workaround.

The issue that I see, is that with the current state of plugin managers,
which treat your plugins as a set, assume plugins can be composed.
There is no way to specify priorities or when some plugin must be
enabled, which would be required to ensure that the theme is not loaded
or that the theme override is ran after loading the theme.
Let's assume that this "right" way of fixing this is just too expensive for
fixing a single ssh program.


What happens today with `fisher`, is that `~/.config/fish/functions/fish_prompt.fish`
and a few others are symlinks to the theme plugin.
Files under that directory are read again over time, so any "dumb-term-fix"
plugin that clears the prompt for me on `~/.config/fish/config.d/` would work for
at most for a few prompt invocations.

To allow theme plugins not to care about the current `TERM` set, then we
would need some "theme-muxer" plugin that is able to load the right prompt
on the right situation, for instance switch between "bobthefish" and "dumbfish"
prompts as needed.
That would require that neither "bobthefish" or "dumbfish" write to the prompt
function directly. Only the "theme-muxer" plugin can do that, and additionally
it needs to know which prompt plugins exist and which one to use.
I don't know how to achieve that easily, it requires coordination between all
theme plugins and some way to rank plugins.

On the other side, this hack allows you to just add bobthefish to your plugin
set and not care about it anymore.

Dietr1ch:
Hey, can you reconsider merging that or close this as a won't fix?

PR #172, opened 2018-10-10
fix VirtualEnv nerdfonts character in PyCharm

andrewgazelka:
Previous to this, the character was not displaying. However, it worked in iTerm2. Also, it could be seen in IntelliJ/PyCharm if copy-pasted but not under the blue prompt background.

**Before:**
image

**After:**
image


yozlet:
I can't speak for PyCharm, but this PR works for me in VSCode terminal. 👍

bobthecow:
The problem with \uE606 is that it's awkwardly large, while \uE73C fits perfectly 😕

Are PyCharm and VS Code using the exact same font as your terminal?

andrewgazelka:
This isn't going to be merged in, and there's a merge conflict, so I'm closing this PR.

Issue #187, opened 2019-01-22
git prompt - number of stashes


lorthirk:
Hey,

Can we have number of current stashes in the git prompt? If it helps, Powerline9k [supports](https://github.com/bhilburn/powerlevel9k#vcs) it :)

EDIT: A way to count them could be `git rev-list --walk-reflogs --count refs/stash`

bobthecow:
This is now supported. Thanks :)

PR #188, opened 2019-01-22
Add Stashes count to Git Prompt

lorthirk:
Hi,

This PR aims to implement #187. This is my first attempt in writing scripts in Fish (well, sh scripts in general 😄) so let me know if I have to fix something!

It introduces a new variable, `$theme_display_git_stashed_verbose`, which adds the stashes count to the prompt if set to `yes`

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -285,6 +285,17 @@ function __bobthefish_git_dirty_verbose -S -d 'Print a more verbose dirty state 
     echo "$changes " | string replace -r '(\+0/(-0)?|/-0)' ''
 end
 
+function __bobthefish_git_stashed -S -d 'Print the stashed state for the current branch'
+    set -l stashes (command git rev-list --walk-reflogs --count refs/stash 2>/dev/null)
+    or return
+
+    echo -n "$git_stashed_glyph"
+    if [ "$theme_display_git_stashed_verbose" = 'yes' ]
+        echo -n $stashes
+    end
+
+end
+
 
 # ==============================
 # Segment functions
@@ -828,7 +839,7 @@ function __bobthefish_prompt_git -S -a git_root_dir -d 'Display the actual git s
     end
 
     set -l staged (command git diff --cached --no-ext-diff --quiet --exit-code 2>/dev/null; or echo -n "$git_staged_glyph")
-    set -l stashed (command git rev-parse --verify --quiet refs/stash >/dev/null; and echo -n "$git_stashed_glyph")
+    set -l stashed (__bobthefish_git_stashed)
     set -l ahead (__bobthefish_git_ahead)
 
     set -l new ''
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -286,12 +286,12 @@ function __bobthefish_git_dirty_verbose -S -d 'Print a more verbose dirty state 
 end
 
 function __bobthefish_git_stashed -S -d 'Print the stashed state for the current branch'
-    set -l stashes (command git rev-list --walk-reflogs --count refs/stash 2>/dev/null)
+    command git rev-parse --verify --quiet refs/stash >/dev/null
     or return
 
     echo -n "$git_stashed_glyph"
     if [ "$theme_display_git_stashed_verbose" = 'yes' ]
-        echo -n $stashes
+        echo -n (command git rev-list --walk-reflogs --count refs/stash 2>/dev/null)
     end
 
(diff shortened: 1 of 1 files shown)


lorthirk:
Sorry about the branch mess... I had to push force a couple time and got lost in the process 😄 Everything should be fine now!

fish_prompt.fish
```diff
@@ -285,6 +285,17 @@ function __bobthefish_git_dirty_verbose -S -d 'Print a more verbose dirty state
     echo "$changes " | string replace -r '(\+0/(-0)?|/-0)' ''
 end
 
+function __bobthefish_git_stashed -S -d 'Print the stashed state for the current branch'
+    set -l stashes (command git rev-list --walk-reflogs --count refs/stash 2>/dev/null)
```
bobthecow: Counting stashes for users who haven’t explicitly asked for it is probably too much of a performance hit to take. Have you profiled this against the current stash flag implementation?

lorthirk:
I made a little tweak in 69516c8. Now stashes are counted only if the variable is set to `yes`.

fish_prompt.fish
lorthirk: I made a little tweak in 69516c8. Now stashes are counted only if the variable is set to `yes`.

fish_prompt.fish
```diff
@@ -285,6 +285,17 @@ function __bobthefish_git_dirty_verbose -S -d 'Print a more verbose dirty state
     echo "$changes " | string replace -r '(\+0/(-0)?|/-0)' ''
 end
 
+function __bobthefish_git_stashed -S -d 'Print the stashed state for the current branch'
+    command git rev-parse --verify --quiet refs/stash >/dev/null
+    or return
+
+    echo -n "$git_stashed_glyph"
+    if [ "$theme_display_git_stashed_verbose" = 'yes' ]
+        echo -n (command git rev-list --walk-reflogs --count refs/stash 2>/dev/null)
```
bobthecow: you don't need the the outer `echo -n (...)` here. it's just capturing stdout and re-echoing it :)

bobthecow:
>No, honestly I didn't... How can I do that? I'm sorry, but as I said I'm a newbie in shell scripting.

You can run fish with the `--profile` flag to time things. I just ran something quick and dirty to test them:

```
for i in (seq 100)
    fish --profile $i.prof --command 'command git rev-parse --verify --quiet refs/stash >/dev/null'
end
```

… for both the current and proposed stash flags code, on a repo with 18 things stashed (yeah, i know, one of these days i'll get to them). In this case, we're only interested in the bottommost entry in each profile file, because everything else is just overhead, so:

```
tail -q -n1 *.prof | bcopy
```

… copies all of them, and by pasting into Excel you can get an average, or whatever else you're interested in. In my test case, your stash count code averaged `8134` (units aren't important) and the existing code averaged `5631`, so it's about 30% slower to count vs not.

That said, in the exact same repo but with nothing stashed, the verbose variant averaged `6229`, which isn't unreasonable. So I'd say that we probably don't need to run it _twice_ for people who want a count. We can likely just choose whether to run the count variant or the non-count variant based on `$theme_display_git_stashed_verbose`. For users who want the verbose output, we can save them the time for the initial boolean check.

bobthecow:
(sorry, unintentionally closed)

fish_prompt.fish
lorthirk: You're right of course :smile: The unnecessary `echo` is removed in 40c597a.

lorthirk:
Ok, in f812a27 now the script either executes non-verbose or verbose variant, not both

lorthirk:
In 529ea35 I removed a space after the stash glyph, otherwise it would have shown as this screenshot:

schermata 2019-01-23 alle 13 50 52

However... It's probably a good idea to have some spacing between a counter and the following glyph if someone has multiple verbose items in the prompt. It would look much cleaner IMHO. What do you think?

patrickf3139:
Just want to say I would love to see this as a feature!

bobthecow:
Thanks for this :)

patrickf3139:
We should probably update the README...
I'll see if I can get around to opening a PR for it next week.

bobthecow:
https://github.com/oh-my-fish/theme-bobthefish/commit/da6379491f4cdae5c60bd7ecf5ea79229214d987 :)

patrickf3139:
Wonderful! thank you all

PR #189, opened 2019-01-22
fish_greeting.fish: Make existance of uptime command optional

ZombineDev:
The `uptime` command does not exist in some environments, such as MSYS2,
so execute the command only if it exists.

### DIFF
--- a/fish_greeting.fish
+++ b/fish_greeting.fish
@@ -1,6 +1,6 @@
 function fish_greeting -d "What's up, fish?"
     set_color $fish_color_autosuggestion
     uname -nmsr
-    uptime
+    command -q -s uptime; and uptime
     set_color normal
 end


fish_greeting.fish
```diff
@@ -1,6 +1,6 @@
 function fish_greeting -d "What's up, fish?"
     set_color $fish_color_autosuggestion
     uname -nmsr
-    uptime
+    command -v uptime > /dev/null; and uptime
```
bobthecow: Let’s use `command -q -s` rather than redirecting stdout

ZombineDev: Thanks, that's indeed better. Commit amended.

bobthecow:
Thank you!

ZombineDev:
Thanks for the awesome theme and for working on Fish 3.0 support!

PR #197, opened 2019-03-28
Fix status reporting

nertpinx:
The status needs to be saved before anything else is done, otherwise it will get
overwritten.  Since commit ac45a5cb9594 this was the case with the `if` block in
a way that `last_status` was always `0`.


bobthecow:
Thanks!

Issue #195, opened 2019-03-19
Make trunkating of long branch names optional


jschneider:
This commit added a nice feature to truncate long git names:
https://github.com/oh-my-fish/theme-bobthefish/commit/f49a8424cd11f3e2f29807c099d61b7327a0c152

I personally would like to configure the max length and/or disable the feature completely.

Greg0:
Yep, I agree. I was looking for that option to configure

PR #200, opened 2019-05-13
Make git branch truncate optional

ribugent:
The changes on f49a8424cd11f3e2f29807c099d61b7327a0c152 modified the default branch display on prompt.

This pr restores the default behaviour and add a configuration variable to make it optional.

This may solve #195


bobthecow:
:+1: to making this optional, but we should truncate by default :)

prolly switch to `theme_git_branch_length` (by analogy with `fish_prompt_pwd_dir_length`) and let users specify the max characters before truncating, or set to `0` to disable?

bobthecow:
Sorry! I completely forgot about this pull request, got annoyed with the previous default, and changed it myself 😬

Fixed by cf096612c832b4b3f908200b5915bc695de0b611

PR #204, opened 2019-06-20
Add prompt for Guix environment.

meiyopeng:
This patch adds prompt for [Guix](https://www.gnu.org/software/guix/) environment.

What is Guix environment: https://www.gnu.org/software/guix/manual/en/html_node/Invoking-guix-environment.html
> The purpose of **guix environment** is to assist hackers in creating reproducible development environments without polluting their package profile. The guix environment tool takes one or more packages, builds all of their inputs, and creates a shell environment to use them. 

The value of $GUIX_ENVIRONMENT is too long to fit in the prompt string and it's content is cryptic.  Something like this: `/gnu/store/23lmq99gs7pqz0vlarjwz3bwws0h02jf-profile`.  Therefore this patch only prints "GUIX ENV" to the prompt line to indicate if we are in a Guix environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ set -g theme_display_vagrant yes
 set -g theme_display_docker_machine no
 set -g theme_display_k8s_context yes
 set -g theme_display_hg yes
+set -g theme_display_guix_environment no
 set -g theme_display_virtualenv no
 set -g theme_display_ruby no
 set -g theme_display_user ssh
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -29,6 +29,7 @@
 #     set -g theme_display_docker_machine no
 #     set -g theme_display_k8s_context yes
 #     set -g theme_display_hg yes
+#     set -g theme_display_guix_environment no
 #     set -g theme_display_virtualenv no
 #     set -g theme_display_ruby no
 #     set -g theme_display_user ssh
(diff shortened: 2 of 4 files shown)


bobthecow:
if guix is run with `-r` it _does_ have a non-cryptic, useful name, right? should we use basename of that path or something for those cases?

i'm not a huge fan of `GUIX ENV ` in the prompt. it's quite loud compared to the rest of the segments. maybe just `guix`, or an applicable glyph from nerd fonts if available?

meiyopeng:
Hi @bobthecow,

Thank you for your feedback.

> if guix is run with `-r` it _does_ have a non-cryptic, useful name, right?

No.  `-r` option creates a symlink to `$GUIX_ENVIRONMENT` on file system, but the content of `$GUIX_ENVIRONMENT` is still cryptic.

> i'm not a huge fan of `GUIX ENV ` in the prompt. it's quite loud compared to the rest of the segments. maybe just `guix`, or an applicable glyph from nerd fonts if available?

`Guix` is too broad.  It can stand for the [GNU Guix](https://www.gnu.org/software/guix/) distro(the operating system), or the GNU Guix functional package manager.  `Guix environment` is only a small use case of the Guix functional package manager.  Thus I don't think `Guix` can properly represent `guix environment`.  I agree that `GUIX ENV` does not look good in the prompt, but I cannot come up with a better choice.  What do you think?

pmeiyu:
`guix environment` has been replaced by `guix shell`. So this patch is updated accordingly. The new prompt string is now a simple "guix".

PR #206, opened 2019-06-20
Add prompt for nix shell.

dukejones:
Show whether there is a current nix-shell environment open.

Uses the nerd-font nix glyph if available.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -801,6 +801,20 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
     set_color normal
 end
 
+function __bobthefish_prompt_nix -S -d 'Display current nix environment'
+    [ "$theme_display_nix" = 'no' -o -z "$IN_NIX_SHELL" ]
+    and return
+
+    if [ "$IN_NIX_SHELL" = "impure" ]
+        set say_what "nix"
+    else
+        set say_what "pure"
+    end
+
+    __bobthefish_start_segment $color_nix
+    echo -ns $nix_glyph' ' "$say_what" ' '
+    set_color normal
+end
 
 # ==============================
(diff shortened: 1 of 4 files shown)


fish_prompt.fish
```diff
@@ -801,6 +802,20 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
 end
 
+function __bobthefish_prompt_nix -S -d 'Display current nix environment'
+    [ "$theme_display_nix" = 'no' -o -z "$IN_NIX_SHELL" ]
+    and return
+
+    if [ "$IN_NIX_SHELL" = "impure" ]
+        set say_what "nix"
+    else
+        set say_what "pure"
```
bobthecow: bobthefish biases towards removing info when things are in the default state, and only showing them when they're "interesting". i don't know enough about nix to know whether this is a case like that, but it seems like it might be. should we be leaving off either the "pure" or "impure" side?

fish_prompt.fish
```diff
@@ -801,6 +802,20 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
+    end
+
+    __bobthefish_start_segment $color_nix
+    echo -ns $nix_glyph' ' "$say_what" ' '
```
bobthecow: normally we put the trailing space on the glyph itself, when the glyph is optional (e.g. when it's from geek fonts). otherwise we end up with awkward spacing when the glyph is removed.

hansl:
I was about to make a similar PR; I use bobthefish and nix and would really like to have those two interact better.

@dukejones @bobthecow What's the status of this PR? Can I help in any way?

bobthecow:
Merged in #224 :)

PR #207, opened 2019-06-26
Add NVM prompt

### LINKED ISSUE #97: Add NVM support

pioneerskies:
As it says. Should fix #97 

Sincerly I'm not 100% sure about colors I've used for all the themes: I chose the green and applied it down to all the supported color schemes.

Since this is a PR on a project I do not maintain, I decided to have this prompt disabled by default.

If NVM version is deactivated (`nvm deactivate`) the nvm prompt won't show. I'm open to other/smarter methods to understand if we're not on an nvm's node version.

Documentation should be ok also.

Note: this is a specific PR done for NVM, hence it's not meant to supersede the (stalled?) #128

### DIFF
--- a/functions/__bobthefish_glyphs.fish
+++ b/functions/__bobthefish_glyphs.fish
@@ -62,6 +62,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set virtualenv_glyph \uE73C ' '
     set ruby_glyph       \uE791 ' '
     set go_glyph         \uE626 ' '
+    set node_glyph       \uF898 ' '
 
     set vagrant_running_glyph  \uF431 # ↑ 'running'
     set vagrant_poweroff_glyph \uF433 # ↓ 'poweroff'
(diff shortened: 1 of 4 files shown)


pioneerskies:
Sry for the push-force, but I've rebased after a fixup in order to retain clean history: I left a debug `echo` in the code. My bad.

pioneerskies:
I've opted to not show the prompt also if `system` version is used

functions/__bobthefish_glyphs.fish
```diff
@@ -62,6 +62,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set virtualenv_glyph \uE73C ' '
     set ruby_glyph       \uE791 ' '
     set go_glyph         \uE626 ' '
+    set node_glyph       \uF898 ' '
```
bobthecow: let's use `\uE718` instead? they look like they're the same, but `\uE718` has been in the fonts for a lot longer, so there's a better chance that existing users will have it in their patched fonts.

pioneerskies: Sure. Was not aware of the fact; thank you.

bobthecow:
Thank you!

PR #208, opened 2019-06-27
Added the k8s namespace value to the prompt and the k8s icon to the segment

### LINKED ISSUE #198: [Feature request] Add togglable kube namespace to prompt
Currently only current context is displayed in prompt when `theme_display_k8s_context' is set to `yes`.

Please consider adding:
- namespace value to prompt;
- add possibility to quickly toggle kube content:namespace globally and for session to see this only when working with Kubernetes cluster.

Functionally similar example containing above features for bash and zsh:
 https://github.com/jonmosco/kube-ps1

morozov:
1. The current Kubernetes namespace is displayed if the current context is not empty and defaults to "default".
2. The `⎈` glyph has been added to the k8s prompt segment.

Fixes #198.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -579,10 +579,7 @@ function __bobthefish_prompt_docker -S -d 'Display Docker machine name'
     echo -ns $DOCKER_MACHINE_NAME ' '
 end
 
-function __bobthefish_prompt_k8s_context -S -d 'Show current Kubernetes context'
-    [ "$theme_display_k8s_context" = 'yes' ]
-    or return
-
+function __bobthefish_k8s_context -S -d 'Get the current k8s context'
     set -l config_paths "$HOME/.kube/config"
     [ -n "$KUBECONFIG" ]
     and set config_paths (string split ':' "$KUBECONFIG") $config_paths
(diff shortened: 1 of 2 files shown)


fish_prompt.fish
```diff
@@ -594,14 +591,39 @@ function __bobthefish_prompt_k8s_context -S -d 'Show current Kubernetes context'
...
     end
+
+    return 1
+end
+
+function __bobthefish_k8s_namespace -S -d 'Get the current k8s namespace'
+    set -l namespace (kubectl config view --minify --output "jsonpath={..namespace}")
+
+    if [ -z "$namespace" ]
+        echo "default"
```
bobthecow: bobthefish tends to omit "default" or "system" whenever possible to keep things clean and only show interesting information. for example, the `master` branch name is omitted when looking at a git repo. does it make sense to leave `default` off here?

fish_prompt.fish
```diff
@@ -594,14 +591,39 @@ function __bobthefish_prompt_k8s_context -S -d 'Show current Kubernetes context'
             if [ "$key" = 'current-context:' ]
                 set -l context (string trim -c '"\' ' -- $val)
                 [ -z "$context" ]
-                and return
+                and return 1
```
bobthecow: what does context `1` mean?

fish_prompt.fish
morozov: Yes. Let's omit `default`.

fish_prompt.fish
morozov: It's not the value of the context, it's a non-zero return status of the function meaning that the context is not detected, in this case, the calling function will not use the output of the function as the context and detect the namespace.

bobthecow: Right. Sorry, I completely missed that :)

morozov:
@bobthecow pleas see the update. Now it displays the k8s as following:

- Non-default namespace:
   ```
   ⎈ some-cluster:some-namespace
   ```
- Default namespace:
   ```
   ⎈ minikube
   ```

bobthecow:
Thanks!

PR #213, opened 2019-10-10
Add option to display the number of background jobs

FMeinicke:
This adds another option (`theme_show_job_number`) that can be set to `yes` in the `config.fish` to display the number of background jobs next to the percent sign.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ set -g theme_avoid_ambiguous_glyphs yes
 set -g theme_powerline_fonts no
 set -g theme_nerd_fonts yes
 set -g theme_show_exit_status yes
-set -g theme_show_job_number yes
+set -g theme_display_jobs_verbose yes
 set -g default_user your_normal_user
 set -g theme_color_scheme dark
 set -g fish_prompt_pwd_dir_length 0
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -98,6 +98,7 @@ set -g theme_avoid_ambiguous_glyphs yes
 set -g theme_powerline_fonts no
 set -g theme_nerd_fonts yes
 set -g theme_show_exit_status yes
+set -g theme_show_job_number yes
```
bobthecow: `theme_display_jobs_verbose` would probably be more consistent with the other configuration options. `theme_show_exit_status` is the outlier :P

fish_prompt.fish
```diff
@@ -449,7 +451,11 @@ function __bobthefish_prompt_status -S -a last_status -d 'Display flags for a no
         if [ "$bg_jobs" ]
             set_color normal
             set_color -b $color_initial_segment_jobs
-            echo -n $bg_job_glyph
+            if [ "$theme_show_job_number" = 'yes' ]
+                echo -ns $bg_job_glyph $bg_jobs ' '
```
bobthecow: you called this variable `$num_bg_jobs` :P 

… but with my change above, we can leave it `$bg_jobs`.

fish_prompt.fish
```diff
@@ -420,7 +421,8 @@ function __bobthefish_prompt_status -S -a last_status -d 'Display flags for a no
     and set superuser 1
 
     # Jobs display
-    jobs -p >/dev/null
+    set num_bg_jobs (jobs -p | wc -l)
```
bobthecow: it would be great if we didn't pay the extra forking cost when the config is disabled. something like...

```fish
if [ "$theme_display_jobs_verbose" = 'yes' ]
  set bg_jobs (jobs -p | wc -l)
  [ "$bg_jobs" -eq 0 ]
  and set bg_jobs # clear it out so it doesn't show when `0`
else
  jobs -p >/dev/null
  and set bg_jobs 1
end
```

README.md
FMeinicke: Will do. Didn't know that `theme_show_exit_status` was the outlier :D

fish_prompt.fish
FMeinicke: I'll use your version with the above code so that we can leave it `$bg_jobs`

PR #215, opened 2019-10-19
Use `command uptime` to avoid coloured output when using grc plugin

FMeinicke:
When using the generic colouriser plugin the output of `uptime` would be coloured. Since I think it is not the intention to have a coloured greeting, using `command uptime` prevents the overriding of the colourisation by `grc`.


bobthecow:
Thanks!

PR #216, opened 2019-10-21
Add option to display sudo-username.

FMeinicke:
This adds another config option `theme_display_sudo_user`. When set to `yes` it displays the content of `$SUDO_USER` in the username segment of the prompt. 

This means that if a user called `normal_user` calls `sudo -s`, for example, and has this option and `theme_display_user` set to yes, the prompt would display `root (normal_user)` in the username segment.  
If `theme_display_user` is not set to `yes` but `theme_display_sudo_user` is, the prompt would display just `(normal_user)` in the username segment.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -692,6 +692,7 @@ function __bobthefish_prompt_user -S -d 'Display current user and hostname'
     end
 
     set -q display_user
+    or set -q display_sudo_user
     or set -q display_hostname
     and echo -ns ' '
 end
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -33,6 +33,7 @@
 #     set -g theme_display_ruby no
 #     set -g theme_display_user ssh
 #     set -g theme_display_hostname ssh
+#     set -g theme_display_sudo_user yes
 #     set -g theme_display_vi no
 #     set -g theme_display_nvm yes
 #     set -g theme_avoid_ambiguous_glyphs yes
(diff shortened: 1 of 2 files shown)


fish_prompt.fish
```diff
@@ -665,6 +669,15 @@ function __bobthefish_prompt_user -S -d 'Display current user and hostname'
         echo -ns (whoami)
     end
 
+    if set -q display_sudo_user
```
bobthecow: We need to update the condition for showing an extra space on line 695. It looks great if you have `theme_display_user` too, but not when it’s only `theme_display_sudo_user`.

FMeinicke: I remember having this condition but while playing around I must have deleted it... :man_facepalming:  
Anyway, I added it back in and it should work now as expected.

fish_prompt.fish
```diff
@@ -665,6 +669,15 @@ function __bobthefish_prompt_user -S -d 'Display current user and hostname'
...
+        if set -q display_user
+            echo -ns ' '
+        else
+            __bobthefish_start_segment $color_username
+        end
+        echo -ns "($SUDO_USER)"
+    end
+
     if set -q display_hostname
         if set -q display_user
```
bobthecow: Oh, I missed this. It will need to be updated to not start a new segment if _either_ `display_sudo_user` or `display_user` is true.

FMeinicke: Oh, you're totally right! I missed that, too.  
Fixed that now!

bobthecow:
Thanks!

PR #217, opened 2019-10-26
README: Clarify git_worktree_support default

andys8:
`theme_git_worktree_support` is enabled by default. The wording "leave it disabled" suggests the option would be disabled.

image
https://github.com/oh-my-fish/theme-bobthefish/blob/master/README.md#configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ set -g theme_newline_prompt '$ '
 - `theme_display_hostname`. Same behaviour as `theme_display_user`.
 - `theme_show_exit_status`. Set this option to `yes` to have the prompt show the last exit code if it was non_zero instead of just the exclamation mark.
 - `theme_display_jobs_verbose`. If set to `yes` this option displays the number of currently running background jobs next to the percent sign.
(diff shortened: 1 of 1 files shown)


bobthecow:
It is disabled by default:

https://github.com/oh-my-fish/theme-bobthefish/blob/master/fish_prompt.fish#L139

You have to set it to `yes` for that conditional code to run.

andys8:
Ah, okay.

I read the paragraph in the readme that way.

> You can override some of the following default options in your config.fish

So should the value be changed to `no` in the configuration list? Or is the list not meant to show the defaults?

bobthecow:
The list is meant to show overrides, since setting defaults doesn’t really make sense. But I can see how it’s confusing :)

andys8:
Okay :) The word "default" in "override some of the following default options" is kinda confusing. Would be great if they would match the defaults because how else would anybody know besides digging into the code. And maybe it's possible to change the wording of the Readme a bit.

I'll close the PR because the change wouldn't make sense. Thanks for clarification.

PR #221, opened 2019-10-28
improve $HOME detection for superuser on macOS (fixes #219)

### LINKED ISSUE #219: __bobthefish_pwd should use $PWD only
On macOS, root's home directory is /var/root but /var is a symlink to /private/var. 

Consequently, when the current directory is '/var/root', invoking `pwd -P` yields '/private/var/root`.

Since __bobthefish_pwd uses `pwd -P`, bobthefish's prompt does not display '~' for root when in the home directory.

A quick (but perhaps not ideal) solution is to use $PWD only rather than invoking `pwd -P`.

LucaFilipozzi:
On macOS, root's $HOME is '/var/root' but '/var' is a symlink to '/var/private'. Consequently, $HOME ('/var/'root') never matches a normalized $PWD ('/var/private/root') and '~' is never displayed.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -359,7 +359,7 @@ function __bobthefish_path_segment -S -a segment_dir -d 'Display a shortened for
     switch "$segment_dir"
         case /
             set directory '/'
-        case "$HOME"
+        case (builtin realpath "$HOME")
             set directory '~'
         case '*'
             set parent (__bobthefish_pretty_parent "$segment_dir")


fish_prompt.fish
```diff
@@ -359,7 +359,7 @@ function __bobthefish_path_segment -S -a segment_dir -d 'Display a shortened for
     switch "$segment_dir"
         case /
             set directory '/'
-        case "$HOME"
+        case (builtin realpath "$HOME")
```
bobthecow: let's do `case "$HOME" (builtin realpath "$HOME")`. the string comparison should be enough cheaper than a filesystem lookup that it's probably worth doing.

LucaFilipozzi:
Cosing this pull request as I've reworked the `__bobthefish_git_project_dir` and `__bobthefish_hg_project_dir` functions. I'll generate a new pull request.

PR #225, opened 2019-11-13
Use repo root when looking for untracked files

evndean:
Currently, if you add a new file to a subdirectory in a git repo then cd to a sibling directory, the untracked glyph will not appear. This is inconsistent with the behavior of the other git colors and flags, e.g. if you modify a file and cd into a sibling directory, the prompt will still be red and the dirty working directory glyph will still appear.

Here's an example of the current behavior:
untracked-glyph-missing

This PR modifies the query used to find untracked files so it searches from the repo root instead of the current directory.

Here's the same directory with these changes applied:
untracked-glyph-fixed


bobthecow:
I'm incredibly sorry this took so long, but thanks! Merged!

PR #227, opened 2019-11-21
Add OS type detection and custom coloring for root user

bokysan:
This commit includes these new features:

- *A segment which displays OS type.* This is especially useful
  if constantly logging into different computers with different
  systems and you need to know if you need to type `apk`, `apt`,
  `dnf`, or `yum`, for example. Enable it by setting the
  `theme_display_currentos` variable to `yes`.
- *Padlock for SSH connection.* You may enable the padlock by
  setting the `theme_display_ssh_keylock` to `yes`. Padlock is
  shown only when accessing the terminal via a SSH session.
  The padlock is displayed _before_ the hostname.
- *Split username and hostname segments.* This enable each segment
  to be coloured with a different background. Set
  `theme_display_split_user_hostname` to `yes` to enable this
  feature.
- *Color root username differently.* When logged in as a `root`
  user, background (or the color of the username) will change
  to red. New colour has been added to the theme(s) for this
  feature.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -402,6 +402,61 @@ end
 # Status and input mode segments
 # ==============================
 
+function __bobthefish_prompt_currentos -S -d 'Display current OS version'
+    [ "$theme_display_currentos" = 'yes' ]
+    and set -l display_currentos
+    
+    if set -q display_currentos
+
+        if ! set -q __bobthefish_os_type
+            if uname | grep -qi Darwin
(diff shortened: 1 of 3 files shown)


bobthecow:
so many changes at once! it's a lot easier to discuss them separately :)

bokysan:
Kind of started fixing and ended up with a larger patch. :)

Even so, I don't think it's a *lot* of code changes. If you want, we can still split it, though, into separate patches. Any suggestions?

PR #233, opened 2020-01-17
In the prompt, show user info before everything else

morozov:
Logically, if you're on a remote machine, all context information including k8s context and so on belongs to that machine since the SSH session is running on that machine. Therefore, the current user and hostname should be displayed first.


bobthecow:
Yeah, that makes sense.

PR #240, opened 2020-02-19
Simple right prompt segment

brodock:
implement a simple version of the right prompt segment (using the "bullet approach")

right prompt

It is better if used with: 

```
set -g theme_date_format "+%H:%M:%S"
```


bobthecow:
I've used this for several days, and I don't think it's for me. The right prompt is intentionally much more subtle than the left, and adding a segment to it makes it heavier than I'd like.

That said, you can definitely implement this yourself! If you move this file to `$HOME/.config/fish/functions/`, you'll get your custom right prompt rather than the default.

PR #244, opened 2020-03-17
Escape project_root_dir before using it in regex

### LINKED ISSUE #243: improper handling of paths containing parenthesis
When navigating to a git repo who's path contains a parenthesis, the git prompt incorrectly shows the full path to the right of the git symbol (rather than just the subdirectories below the repo root).

bobthefish_parens_test

**Some background:**
Most of my git repos are in Dropbox, and because I have both a work and a personal Dropbox account, Dropbox (annoyingly, and irrevocably :) ) enforces a root dropbox directory name which contains both a space and parenthesis (specifically, `/Users/eric/Dropbox (Personal)/`).

For a while I was getting around this problem by using a symbolic link (which contained no spaces or parens), but today I upgraded to the current bobthefish hoping maybe the parens issue had been fixed, and it actually got worse because the current version ignores the symbolic link and uses the "real" path, so now I see the parens problem even when I use my "workaround" and cd to the symbolic link.

**Version:**
OSX 10.14.6
Not sure how to get my bobthefish version, but I ran `omf update bobthefish` today so presumably it is "current".

epmoyer:
Fixes issue where paths containing parentheses would be shown **in ful**l to the right of the git "project root" symbol instead of showing only the **current subdirectory**.

Fixes #243

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -231,7 +231,8 @@ function __bobthefish_project_pwd -S -a project_root_dir -a real_pwd -d 'Print t
     set -q theme_project_dir_length
     or set -l theme_project_dir_length 0
 
-    set -l project_dir (string replace -r '^'"$project_root_dir"'($|/)' '' $real_pwd)
+    set -l project_root_dir_escaped (string escape --style=regex $project_root_dir)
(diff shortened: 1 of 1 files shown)


bobthecow:
Thank you!

PR #250, opened 2020-04-29
Fix always displaying system ruby when set by asdf

patrickf3139:
Since the asdf-ruby integration with bobthefish was written, asdf has changed their ruby version provenance format.
`asdf current ruby 2>/dev/null | read -l asdf_ruby_version asdf_provenance` now results in 
```
$ echo $asdf_ruby_version
2.5.7
$ echo $asdf_provenance                                                                          
   (set by /Users/patrick.fong/.tool-versions)
```
Notice the extraneous 3 spaces in the beginning. The original asdf integration didn't expect that whitespace and so right now it's treating it as a non-system Ruby. And that results in the system ruby version always being displayed, which is annoying. 

This is a small patch to fix that. All I did was pipe the output through `tr -s` which removes the extra spaces. I've tested it locally to make sure it works. Thanks!

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -781,7 +781,7 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
         or return
 
         # If asdf changes their ruby version provenance format, update this to match
-        [ "$asdf_provenance" = "(set by $HOME/.tool-versions)" ]
+        [ "$asdf_provenance" = "   (set by $HOME/.tool-versions)" ]
         and return
 
         set ruby_version $asdf_ruby_version
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -777,11 +777,11 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
     else if type -q chruby # chruby is implemented as a function, so omitting the -f is intentional
         set ruby_version $RUBY_VERSION
     else if type -fq asdf
-        asdf current ruby 2>/dev/null | read -l asdf_ruby_version asdf_provenance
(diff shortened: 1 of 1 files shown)


bobthecow:
I like the trimming approach. Let’s use the `string` builtin rather than `tr`.

patrickf3139:
Oh wait don't merge yet. That's not gonna work. Hmmm....

patrickf3139:
Ok it's good to go now. Thanks for reviewing and for maintaining this awesome theme, Justin!

patrickf3139:
Justin?

fish_prompt.fish
```diff
@@ -777,7 +777,7 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
     else if type -q chruby # chruby is implemented as a function, so omitting the -f is intentional
         set ruby_version $RUBY_VERSION
     else if type -fq asdf
-        asdf current ruby 2>/dev/null | read -l asdf_ruby_version asdf_provenance
+        asdf current ruby 2>/dev/null | read -l -t asdf_ruby_version asdf_provenance
```
bobthecow, line 780: I'm curious why you didn't go with `string trim`?

Changing `read` to interpret the asdf output as escaped shell strings will likely bite us later.

bobthecow: Oh, checked the output and I understand why trim wouldn't have worked. Lemme see if I can come up with something better.

bobthecow:
Thanks for raising this, @patrickf3139! I ended up adding the trim just as we check the output (see 7c2d90152eb6a2f319c3927fa11adb7763f2ea6f).

Interestingly, in my current fish and adsf versions, it doesn't have extra whitespace in the variable name. But trimming it won't hurt :)

PR #252, opened 2020-05-04
Do not display Kubernetes context and namespace if it's the defaults ones

Rykian:
Kubernetes context is always displayed even if we're using local Kubernetes instance and that's a little annoying :)


bobthecow:
That makes sense. Thanks!

PR #262, opened 2020-08-14
Move Kubernetes prompt color into a function

morozov:
This patch extracts the logic of determining the Kubernetes prompt color to a separate function that a user can override based on their cluster naming.

For example, the following could be added to `~/.config/fish/config.fish`:
```fish
function __bobthefish_prompt_k8s_color -S
    set -l color $color_k8s

    switch $argv[1]
    case "*prod"
        set color red white --bold
    case "*stage"
        set color yellow white --bold
    end

    for i in $color
        echo $i
    end
end
```
It's required that the default implementation of the `__bobthefish_prompt_k8s_color` function is declared in its own file and is autoloadable. This way, the user-defined function, if exists, will be declared first and prevent autoloading of the default implementation.

### DIFF
--- a/functions/__bobthefish_prompt_k8s_color.fish
+++ b/functions/__bobthefish_prompt_k8s_color.fish
@@ -1,0 +1,5 @@
+function __bobthefish_prompt_k8s_color -S -d 'Determine Kubernetes prompt color based on the current context and namespace'
+    for i in $color_k8s
+        echo $i
+    end
+end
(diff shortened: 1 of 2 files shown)


morozov:
@bobthecow the changes in the PR are eventually ready for review. Could you take a look, please?

morozov:
I see that a few kubectl color-related patches were accepted recently. Could you consider this one as well?

PR #264, opened 2020-08-30
Display screen session if you set theme_display_screen to yes

pfsmorigo:
I use multiple screen sessions so I would like to see which one I am in the current terminal.


functions/__bobthefish_colors.fish
```diff
@@ -29,6 +29,7 @@ function __bobthefish_colors -S -a color_scheme -d 'Define colors used by bobthe
       set -x color_k8s                      magenta white --bold
       set -x color_username                 white black --bold
       set -x color_hostname                 white black
+      set -x color_screen                   brgreen $colorfg
```
bobthecow, line 32: can you add an example segment to `bobthefish_display_colors`, then use that to upload screenshots of how it looks in all the themes?

fish_prompt.fish
```diff
@@ -911,6 +912,18 @@ function __bobthefish_prompt_hg -S -a hg_root_dir -a real_pwd -d 'Display the ac
...
 
+function __bobthefish_prompt_screen -S -d 'Display the screen name'
+    [ "$theme_display_screen" != 'yes' ]
+    and return
+
+    [ -z "$STY" ]
+    and return
+
+    __bobthefish_start_segment $color_screen
+    echo -ns (command echo $STY | cut -d. -f2) ' '
```
bobthecow, line 923: we can probably do this entirely with the `string` builtin, saving ourselves a couple of calls out to core utils. what does $STY typically contain?

bobthecow (changes requested):
Awesome. I like the idea, I like making it opt-in rather than opt-out.

Is there an analog the other settings which allow a verbose version, where it'd be useful to know you're in a screen session, but not necessarily which one? I imagine for someone who typically only ran one screen session per machine, that would be preferable.

fish_prompt.fish
pfsmorigo: done

pfsmorigo:
> Awesome. I like the idea, I like making it opt-in rather than opt-out.
> 
> Is there an analog the other settings which allow a verbose version, where it'd be useful to know you're in a screen session, but not necessarily which one? I imagine for someone who typically only ran one screen session per machine, that would be preferable.

Now is opt-in without verbose. I don't know what to show if verbose is off. Now only adds a space with the color. Maybe we can give me an idea.

About the screen name, the format is pid.tty.host: 

```
> screen -list
There are screens on:
        3226255.first  (09/03/2020 08:40:23 AM)        (Attached)
        2824062.second (09/03/2020 05:48:22 AM)        (Attached)
        1901576.third  (08/30/2020 09:18:26 AM)        (Attached)
3 Sockets in /run/screen/S-pfsmorigo.
> echo $STY
3226255.second
```
So I just use the second part (tty) as the screen session.

pfsmorigo:
I don't know if I need to do something else. I think I did all requested changes.

pfsmorigo:
@bobthecow Hello. I think I did all the requested modifications. What do I need to do in order to have this merged? Thanks!

pfsmorigo:
@bobthecow Hello again! Is my pull request ok or should I change something?

pfsmorigo:
Hello, is there something missing from my side? Can we merge this?

pfsmorigo:
@bobthecow hello, can you merge this PR? Please let me know if you want me to change anything.

bobthecow:
@pfsmorigo hi! I'm so sorry, for some reason I missed the pings entirely. I'll take a look!

bobthecow (approved):

bobthecow:
I don't think I have a great thing to do when not verbose, so I'll merge this then remove that option ¯\\\_(ツ)\_/¯

pfsmorigo:
Thank!

PR #266, opened 2020-09-10
Add support for AWS Vault segment in prompt (mk2)

darrenkidd:
This PR adds support for a mature and popular AWS credentials manager called [AWS Vault](https://github.com/99designs/aws-vault). Specifically, it adds a segment with the name of the profile being executed/used and how long you have until the session expires.

Requires that you set the `theme_display_aws_vault_profile` var to `yes` as is the pattern with many other theme vars.

image

**Motivation:**

When assuming a role in AWS using `aws-vault exec some-role` it spawns a new shell. However, it's very easy to forget what role you're in because it looks like any other shell. So if you're bouncing between terminals all day you quickly lose track of what sessions are where and even whether they've expired or not (which can sometimes cause cryptic errors in your tools and scripts).

This PR makes the currently executing profile clearly visible, so you know which shell to use, which reduces the likelihood of costly mistakes such as accidentally deleting things in your PROD environment when you think you're in TEST.

**Colours:**

Here is output of `bobthefish_display_colors --all`.

themes_1
themes_2
themes_3
themes_4
themes_5
themes_6

### DIFF
--- a/functions/bobthefish_display_colors.fish
+++ b/functions/bobthefish_display_colors.fish
@@ -142,7 +142,7 @@ function bobthefish_display_colors -a color_scheme -d 'Print example prompt colo
   __bobthefish_finish_segments
 
   __bobthefish_start_segment $color_aws_vault
-  echo -ns $aws_vault_glyph ' aws-vault '
+  echo -ns aws-vault ' '
   __bobthefish_finish_segments
 
   echo -e "\n"
--- a/functions/bobthefish_display_colors.fish
+++ b/functions/bobthefish_display_colors.fish
@@ -141,5 +141,9 @@ function bobthefish_display_colors -a color_scheme -d 'Print example prompt colo
   echo -ns $desk_glyph desk ' '
   __bobthefish_finish_segments
 
+  __bobthefish_start_segment $color_aws_vault
+  echo -ns $aws_vault_glyph ' aws-vault '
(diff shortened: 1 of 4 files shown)


functions/bobthefish_display_colors.fish
```diff
@@ -141,5 +141,9 @@ function bobthefish_display_colors -a color_scheme -d 'Print example prompt colo
   echo -ns $desk_glyph desk ' '
   __bobthefish_finish_segments
 
+  __bobthefish_start_segment $color_aws_vault
+  echo -ns $aws_vault_glyph ' aws-vault '
```
bobthecow, line 145: There's no more `$aws_vault_glyph` :)

bobthecow (changes requested):
Other than that `$aws_vault_glyph`, this is looking great!

functions/bobthefish_display_colors.fish
darrenkidd: Looking back over my work, I realised that I did a `drop` instead of `squash` for my glyph fix-up commit. Ugh.

darrenkidd:
Hey @bobthecow, anything left for me to do on this PR? Can it be merged?

bobthecow:
Hey @darrenkidd, sorry, I missed that you'd updated it. Thanks for the nudge :)

PR #267, opened 2020-09-12
use system timezone as default

WiredSharp:
When *theme_date_timezone* is not set, date is displayed as UTC date instead of current system timezone.

This update will check if *theme_date_timezone* is set before setting TZ variable.


bobthecow:
Thanks!

I cleaned up the defaults setting in fish_prompt :)

PR #278, opened 2020-10-18
Improve AWS Vault segment with better session time format and expiration warning

darrenkidd:
The AWS-Vault support added recently rendered session times > 60 mins in minutes only i.e. `702m`. Didn't think this was particularly acceptable/usable, but didn't want to derail the last PR (#266), so adding as a separate PR now.

Also, when it's expired, the segment will now change colour to show.

**Colours:**

image
image
image

### DIFF
--- a/functions/bobthefish_display_colors.fish
+++ b/functions/bobthefish_display_colors.fish
@@ -142,11 +142,11 @@ function bobthefish_display_colors -a color_scheme -d 'Print example prompt colo
   __bobthefish_finish_segments
 
   __bobthefish_start_segment $color_aws_vault
-  echo -ns aws-vault active ' '
+  echo -ns aws-vault ' (' active ') '
   __bobthefish_finish_segments
 
   __bobthefish_start_segment $color_aws_vault_expired
-  echo -ns aws-vault expired ' '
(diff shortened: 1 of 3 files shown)


functions/bobthefish_display_colors.fish
```diff
@@ -142,7 +142,11 @@ function bobthefish_display_colors -a color_scheme -d 'Print example prompt colo
   __bobthefish_finish_segments
 
   __bobthefish_start_segment $color_aws_vault
-  echo -ns aws-vault ' '
+  echo -ns aws-vault active ' '
```
bobthecow, line 145: `echo -ns` here eats the space between "aws-vault" and "active" :)

bobthecow (changes requested):

functions/bobthefish_display_colors.fish
darrenkidd: Can't believe I missed that. Fixed:

image

bobthecow (approved):

bobthecow:
Thanks!

darrenkidd:
Thanks, @bobthecow. Just so it's documented...

**Expired:**
image

**Time Format:**
image

Issue #281, opened 2020-11-12
 system time zone is not used by the right prompt


champignoom:
The right prompt shows UTC+0 time by default, ignoring the system time zone.

hundertzwoelf:
Hi, I had the same problem, but I fixed it by explicitly stating the timezone in my fish config.
In my case I added:
`set -g theme_date_timezone Europe/Berlin`

Henriquelay:
Worked for me! Thanks

bobthecow:
This is fixed now! Thanks for your patience.

PR #282, opened 2020-11-12
do not set empty time zone by default

champignoom:
solves https://github.com/oh-my-fish/theme-bobthefish/issues/281

### DIFF
--- a/fish_right_prompt.fish
+++ b/fish_right_prompt.fish
@@ -63,7 +63,8 @@ function __bobthefish_timestamp -S -d 'Show the current timestamp'
     or set -l theme_date_format "+%c"
 
     echo -n ' '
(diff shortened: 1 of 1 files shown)


FMeinicke:
I think that is already addressed by #267. But it hasn't been merged yet...

champignoom:
> I think that is already addressed by #267. But it hasn't been merged yet...

Indeed. I had only searched for issues before writing the patch.

But still, I hope this less aggressive and more dry patch would not be ignored like the previous one.

bobthecow:
Hey, thanks for nudging on this, and for the smaller PR 😄

Closing in favor of the earlier fix in #267 (I ended up merged just the tz change, and reverting the rest).

PR #283, opened 2020-11-18
Add support for OCI containers (podman)

jkonecny12:
This is an rootless alternative to docker container tool. It has a bunch of benefits. For more information see:
https://podman.io/

The main reason why to have this support is `toolbox` tool which is used as a primary tool for development on immutable system (not-only) such as Fedora SilverBlue. The immutability of the system is replaced by creating containers for projects and work inside these containers. For more info see:
https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/

It's really nice to have this feature here because toolbox will use configuration of the system in the container as it is in host so it's pretty reasonable to use fish inside the container but know that I'm in the container right now.


jkonecny12:
Ping, could this be merged please?

jkonecny12:
Ping, I would like to find out if I have a chance to get the PR merged?

jkonecny12:
Rebased to solve conflicts and solve typo in the variable `$containers` should be `$container`.

jkonecny12:
I'm thinking about creating a fork because this doesn't look like it would be merged ever.

PR #286, opened 2020-12-09
Add FNM as an alternative to NVM

wesbaker:
I use [Schniz/fnm](https://github.com/Schniz/fnm) instead of nvm and wanted the prompt support for the current version. The commands and such are basically the same, so I duplicated it for now, but could see a single config that checked for either and went went with one over the other.

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -78,7 +78,7 @@ function __bobthefish_escape_regex -a str -d 'A backwards-compatible `string esc
 end
 
 function __bobthefish_git_branch -S -d 'Get the current git branch (or commitish)'
-    set -l branch (command git symbolic-ref HEAD | string replace -r '^refs/heads/' '' 2>/dev/null)
+    set -l branch (command git symbolic-ref HEAD 2>/dev/null | string replace -r '^refs/heads/' '')
     and begin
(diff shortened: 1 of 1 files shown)


functions/fish_prompt.fish
```diff
@@ -909,6 +910,20 @@ function __bobthefish_prompt_nvm -S -d 'Display current node version through NVM
     set_color normal
 end
 
+function __bobthefish_prompt_nvm -S -d 'Display current node version through FNM'
```
bobthecow, line 913: this overrides the previously declared NVM function. if we're going to do it this way we need two functions :)

bobthecow (changes requested):
Let's follow the pattern we use for Python and Ruby, where a single segment has a prioritized list of providers that it checks. Since FNM is a drop-in replacement for NVM, it doesn't really make sense to have both in the prompt at the same time.

bobthecow:
Can we split out the "move to functions directory" part to keep this PR focused?

wesbaker:
> Can we split out the "move to functions directory" part to keep this PR focused?

Yeah, that accidentally got included. I'll back that out.

wesbaker:
I'm going to tidy things up and resubmit.

PR #289, opened 2021-01-05
Handle missing generic Python binary gracefully

mzuther:
Hi!

My Linux Mint 20 system doesn't have a generic Python binary (`/usr/bin/python`) anymore. This fix handles this situation gracefully but still prefers an existing generic Python binary.

Happy new year! :)

Martin

### DIFF
--- a/fish_prompt.fish
+++ b/fish_prompt.fish
@@ -847,7 +847,19 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
 end
 
 function __bobthefish_virtualenv_python_version -S -d 'Get current Python version'
-    switch (python --version 2>&1 | tr '\n' ' ')
+    set -l python_version
+
+    # try preferred Python version first
+    if [ (which python) ]
+        set python_version (python --version 2>&1 | tr '\n' ' ')
+    # choose Python 3 over Python 2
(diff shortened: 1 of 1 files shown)


fish_prompt.fish
```diff
@@ -847,7 +847,19 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
 end
 
 function __bobthefish_virtualenv_python_version -S -d 'Get current Python version'
-    switch (python --version 2>&1 | tr '\n' ' ')
+    set -l python_version
+
+    # try preferred Python version first
+    if [ (which python) ]
```
bobthecow, line 853: As a general rule, any `()` in a fish prompt is a performance regression. Let's use the `type` builtin (specifically `type -fq python`) rather than `which` here.

bobthecow (changes requested):

mzuther:
Thanks for the comment!

However, I have found that the missing `python` executable in Linux Mint was not the culprit.

My problem lay elsewhere: a virtualenv environment seems to have broken all of a sudden. On re-creation, a `python` binary was created by virtualenv.

So my PR would hide such a problem and might lead to a wrong display and thus confused users. So I suggest dropping this PR.

PR #292, opened 2021-01-20
Add FNM as an alternative to NVM

wesbaker:
I use [Schniz/fnm](https://github.com/Schniz/fnm) instead of nvm and wanted the prompt support for the current version. The commands and such are basically the same, so I duplicated it for now, but could see a single config that checked for either and went went with one over the other.


fish_prompt.fish
```diff
@@ -895,16 +896,30 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
     set_color normal
 end
 
-function __bobthefish_prompt_nvm -S -d 'Display current node version through NVM'
-    [ "$theme_display_nvm" = 'yes' -a -n "$NVM_DIR" ]
+function __bobthefish_prompt_node -S -d 'Display current node version'
+    [ "$theme_display_node" = "yes" -o "$theme_display_nvm" = 'yes' ]
```
wesbaker, line 900: This supports both `theme_display_node` and `theme_display_nvm` for backwards compatibility.

wesbaker:
@bobthecow I think this takes care of your concerns from #286. Let me know if there are any other issues or backwards compatibilities I need to add.

fish_prompt.fish
```diff
@@ -895,16 +896,30 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
+      set node_manager_dir $NVM_DIR
+    else if type -fq fnm
+      set node_manager "fnm"
+      set node_manager_dir $FNM_DIR
+    end
+
+    [ -n "$node_manager_dir" ]
+    or return
+
+    set -l node_version ("$node_manager" current)
```
bobthecow, line 917: we should swallow errors here, as before (redirect stderr to /dev/null)

fish_prompt.fish
```diff
@@ -895,16 +896,30 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
 
     [ -z $node_version -o "$node_version" = 'none' -o "$node_version" = 'system' ]
     and return
 
-    __bobthefish_start_segment $color_nvm
+    __bobthefish_start_segment $color_node
```
bobthecow, line 922: i know you've updated the default color schemes, but bobthefish also allows custom color schemes. it'd be nice to have a backwards compatible way of supporting this.

maybe in `fish_prompt` and `fish_mode_prompt` where we check `type -q bobthefish_colors`, add a block and after calling their custom colors, check for `$color_nvm` and override `$color_node`?

bobthecow (changes requested):
Great, I like this approach. Two small fixes!

bobthecow:
(this will need a rebase; i've merged the top level .fish -> functions change)

fish_prompt.fish
wesbaker: I'm not sure how to do this in `fish_prompt` or `fish_mode_prompt`, but I can figure it out within `__bobthefish_prompt_node`. I'll push that up. If you'd rather go another route, I'll need your help.

bobthecow: Actually, yeah, it's probably fine to do it here. It's not like anything else needs the node color :)

wesbaker: Never mind. I figured it out.

wesbaker: > Actually, yeah, it's probably fine to do it here. It's not like anything else needs the node color :)

Hmm, good point, I'll opt for that instead. Then it's only one place.

bobthecow (approved):

bobthecow:
Thank you!

PR #293, opened 2021-01-23
Move files to be compatible with Fisher standards

nathan-gilbert:
Fisher 4 doesn't allow any top-level *.fish files in plugins any more. They are standardizing around plugin file being in `functions`. See https://github.com/jorgebucaran/fisher/issues/651

I've tested that the theme works with Fisher after making these changes.


bobthecow:
Thanks!

PR #299, opened 2021-04-13
Create new segment for AWS profile

fernandocarletti:
This change adds the `$AWS_PROFILE` as a new segment in the prompt. I took the previous existing AWS Vault as base and reused the colors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ set -g theme_newline_prompt '$ '
 - `theme_display_vi`. By default the vi mode indicator will be shown if vi or hybrid key bindings are enabled. Use `no` to hide the indicator, or `yes` to show the indicator.
 - `theme_display_k8s_context`. This feature is disabled by default. Use `yes` to show the current kubernetes context (`> kubectl config current-context`).
 - `theme_display_k8s_namespace`. This feature is disabled by default. Use `yes` to show the current kubernetes namespace.
+- `theme_display_aws_profile`. This feature is disabled by default. Use `yes` to show the currently [AWS](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) profile.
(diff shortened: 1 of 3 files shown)


darrenkidd:
Hey @fernandocarletti - nice work!

I authored the AWS Vault segment, and have since thought about doing the AWS profile as well - but you've done the work for me! Cheers!

However, I'm wondering whether we should merge the two together i.e. a generic "AWS segment" that activated on either the vanilla AWS CLI tools or the AWS vault tool. Somewhat prevents crazy config explosion and seems to make sense to keep all the same kinds of things nicely grouped together. Thoughts?

fernandocarletti:
Hey @darrenkidd! If I recall correctly, this was actually my initial approach to this, but I don't remember exactly why I moved away from it. But I agree with you, it makes more sense to have it in a single configuration.

I can change the implementation and get your input once it is ready. Sounds good to you?

darrenkidd:
> _I can change the implementation and get your input once it is ready. Sounds good to you?_

Sounds great!

IMO I think we should prioritise `$AWS_PROFILE` first i.e. use that and then go looking for `$AWS_VAULT` or others if we don't find it. It's the more standard one.

I also think we should move to `$theme_display_aws_profile` as the main var and remove `$theme_display_aws_vault_profile`. For backwards compatibility we'll need to keep the latter around in our prompt function for a while, though.

From memory the AWS CLI also puts session expiry information into env vars too, so we should be able to leverage that now as well? Can rename `color_aws_vault_expired` to `color_aws_expired` then?

fernandocarletti:
@darrenkidd it is finally done! Can you take a look if it fits what you imagined? Best!

darrenkidd:
Nice, @fernandocarletti. I've had a skim but need a bit more time to think and digest. I just want to double-check it handles both tools well with no surprises!

kamilpp:
Hi, any chances to get this into mainstream anytime soon? This is a small change while quite useful, would be great to get it merged :)

PR #305, opened 2021-08-15
Update ruby asdf support for output of asdf v0.8.1

smook1980:
Support for the asdf ruby plugin was broken for me on the latest version of the plugin and asdf installed via brew. 

Changes:
- asdf not output three columns for for `asdf current version`, the first being the plugin name
- Update the provenance format to match


Rykian:
Bumping this, because asdf is not working otherwise.

bobthecow:
Merge! LMK if there are any issues. I don't use adsf so I won't notice.

PR #311, opened 2021-10-25
show git tags if any

zztalker:
Without change tags not shown (because return happen earlier).
With change tags shown if exists.

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -78,6 +78,9 @@ function __bobthefish_escape_regex -a str -d 'A backwards-compatible `string esc
 end
 
 function __bobthefish_git_branch -S -d 'Get the current git branch (or commitish)'
+    set -l tag (command git describe --tags --exact-match 2>/dev/null)
+    and echo "$tag_glyph $tag "
+
     set -l branch (command git symbolic-ref HEAD 2>/dev/null | string replace -r '^refs/heads/' '')
     and begin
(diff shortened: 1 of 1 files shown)


leo150:
bump

zztalker:
@tommyknows @bobthecow Could you take a look?

bobthecow:
Heh. Sorry about that, somehow I completely missed when this was opened 😬

bobthecow:
Okay so it seems like this changes it to always show a tag _and_ either a branch or commit ref. I'm not sure I love this, to be honest. In the cases where you're looking at a tag that's not the current branch head, showing the commit feels like it'll just add unnecessary noise.

zztalker:
> Okay so it seems like this changes it to always show a tag _and_ either a branch or commit ref. I'm not sure I love this, to be honest. In the cases where you're looking at a tag that's not the current branch head, showing the commit feels like it'll just add unnecessary noise.

We can make it a configurable behavior.  As for me, it adds not much noise, but useful. 

Screenshot 2023-06-29 at 11 05 06

bobthecow:
no, that makes sense. 

i'm specifically talking about the case where you check out a previous tag, which isn't a branch HEAD. previously, it would show the tag:

Screenshot 2023-06-29 at 9 14 32 AM

now it shows the tag and the commit:

Screenshot 2023-06-29 at 9 16 26 AM

the latter is just noise, IMO.

PR #314, opened 2021-11-30
Add a rustc version segment with multiple install options

hansl:


hansl:
Show, not just tell:

image

hansl:
@bobthecow This is ready to review when you want, and has been rebased on laster master branch. Thanks!

functions/fish_prompt.fish
```diff
@@ -846,6 +847,47 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+    # If this value is non-zero we should show it as a segment.
+    set -l local_rustc_version
+
+    switch "$theme_display_rustc"
+        case always
+            set local_rustc_version (__bobthefish_get_cargo_version (__bobthefish_pwd))
+        case yes
+            cargo locate-project --message-format plain --quiet >/dev/null 2>/dev/null
+            and set local_rustc_version (__bobthefish_get_cargo_version (__bobthefish_pwd))
+        case diff
```
bobthecow, line 872: This is the default behavior for most segments (e.g. ruby). Can we make two options, always and yes, and have yes do what diff currently does?

(Alternatively, if it's not to expensive to default on, it would be off/always).

functions/__bobthefish_glyphs.fish
```diff
@@ -70,6 +71,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set nix_glyph        \uF313 ' ' # nf-linux-nixos
     set virtualenv_glyph \uE73C ' '
     set ruby_glyph       \uE791 ' '
+    set rustc_glyph       \uE7a8 ' ' # Rust Nerd Font logo
```
bobthecow, line 74: Minor things but you clearly copied the Ruby line, which has one fewer character, for these two and the white space is off. Do you mind lining them all up? :)

functions/fish_prompt.fish
```diff
@@ -846,6 +847,47 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
     echo -ns $ruby_glyph $ruby_version ' '
 end
 
+function __bobthefish_get_cargo_version -S -d 'Display the cargo version in a directory'
+    pushd $argv[1]
+    command rustc --version | cut -d ' ' -f 2
```
bobthecow, line 852: Does rustc take a working directory to save us from a pushd/popd? Otherwise, we should probably add `command`.

bobthecow (commented):
How expensive are cargo lookups? Can we default to on if rustc is present?

functions/fish_prompt.fish
hansl: It takes 70msec to run `rustc --version`. I _think_ it's too much to run with every prompt; if you had 2-3 of these it would be a third of a second and it's noticeable. But I'll defer the final decision to you.

I've done the first part.

functions/fish_prompt.fish
hansl: Unfortunately no. I'll add `command` but I'm also exploring if there's a better way to get this information.

hansl:
> How expensive are cargo lookups? Can we default to on if rustc is present?

image

functions/fish_prompt.fish
hansl: So I decided to rely on `fish -c " ... "` to create a subshell only if we need to (which should be once for the global anyway).

hansl:
PTAL.

hansl:
Ping, @bobthecow

hansl:
@bobthecow Ping. I have additional changes that I would like to submit a PR for, but they're based on this and don't want to pollute your queue.

PR #315, opened 2021-11-30
Add more options to conditionally show node version

hansl:
There are now 4 options for `$theme_display_node`:

- the default is (still) to never show it.
- `yes` to always show the version.
- `rc` to show it if we can find a `.nvmrc` or `.node-version` file in the parent directories.
- `package` to show it if we can find a `package.json` file in the parent directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ set -g theme_newline_prompt '$ '
 **Prompt options**
 
 - `theme_display_ruby`. Use `no` to completely hide all information about Ruby version. By default Ruby version displayed if there is the difference from default settings.
-- `theme_display_node`. If set to `yes`, will display current NVM or FNM node version. If set to `rc`, will display the version if an `.nvmrc` or `.node-version` file is found. If set to `package`, will display the version if a `package.json` file is found.
+- `theme_display_node`. If set to `always`, will display current NPM, NVM or FNM node version. If set to `yes`, will display the version if an `.nvmrc`, `.node-version` or `package.json` file is found in the parent directories.
 - `theme_display_vagrant`. This feature is disabled by default, use `yes` to display Vagrant status in your prompt. Please note that only the VirtualBox and VMWare providers are supported.
 - `theme_display_vi`. By default the vi mode indicator will be shown if vi or hybrid key bindings are enabled. Use `no` to hide the indicator, or `yes` to show the indicator.
 - `theme_display_k8s_context`. This feature is disabled by default. Use `yes` to show the current kubernetes context (`> kubectl config current-context`).
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ set -g theme_newline_prompt '$ '
 **Prompt options**
 
 - `theme_display_ruby`. Use `no` to completely hide all information about Ruby version. By default Ruby version displayed if there is the difference from default settings.
-- `theme_display_node`. If set to `yes`, will display current NVM or FNM node version.
+- `theme_display_node`. If set to `yes`, will display current NVM or FNM node version. If set to `rc`, will display the version if an `.nvmrc` or `.node-version` file is found. If set to `package`, will display the version if a `package.json` file is found.
 - `theme_display_vagrant`. This feature is disabled by default, use `yes` to display Vagrant status in your prompt. Please note that only the VirtualBox and VMWare providers are supported.
 - `theme_display_vi`. By default the vi mode indicator will be shown if vi or hybrid key bindings are enabled. Use `no` to hide the indicator, or `yes` to show the indicator.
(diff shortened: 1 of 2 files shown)


hansl:
Hey @bobthecow , do you think there's anything missing from this PR? I've been using it locally for a while now and it works great. It just toggles whether or not the node version should be shown.

Basically I set it up so that I show only when there's a package.json, e.g.

image

Let me know, thanks!

functions/fish_prompt.fish
```diff
@@ -895,9 +895,42 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
+    if [ "$theme_display_node" = 'yes' -o "$theme_display_nvm" = 'yes' ]
+        set should_show 1
+    else if [ "$theme_display_node" = 'rc' ]
+        begin
+            __bobthefish_prompt_find_file "$PWD" .nvmrc
+            or __bobthefish_prompt_find_file "$PWD" .node-version
+        end
+        and set should_show 1
+    else if [ "$theme_display_node" = 'package' ]
+        __bobthefish_prompt_find_file "$PWD" package.json
```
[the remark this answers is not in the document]
hansl: In this case it doesn't matter. If you have a `package.json` in your history you can always run `npm` so the node version is relevant.

functions/fish_prompt.fish
```diff
@@ -895,9 +895,42 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
     set_color normal
 end
 
+function __bobthefish_prompt_find_file -S -d 'Find a file by going up the parent directories'
```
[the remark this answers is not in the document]
hansl: Decided to go with regex because it's easier with `string match`. WDYT?

README.md
```diff
@@ -124,7 +124,7 @@ set -g theme_newline_prompt '$ '
 **Prompt options**
 
 - `theme_display_ruby`. Use `no` to completely hide all information about Ruby version. By default Ruby version displayed if there is the difference from default settings.
-- `theme_display_node`. If set to `yes`, will display current NVM or FNM node version.
+- `theme_display_node`. If set to `yes`, will display current NVM or FNM node version. If set to `rc`, will display the version if an `.nvmrc` or `.node-version` file is found. If set to `package`, will display the version if a `package.json` file is found.
```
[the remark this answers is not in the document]
hansl: Fine by me, but it'd be a breaking change with the old behavior which I was trying to avoid. I'll do it in a separate commit so it's easy to revert this one comment if you want.

hansl:
@bobthecow Thanks for the review. I addressed all points. PTAL.

I have another similar PR with Rust support that I will update when this one is approved, and ping you on when ready.

Cheers!

functions/fish_prompt.fish
```diff
@@ -895,9 +895,36 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
+function __bobthefish_prompt_find_file_up_re -S -d 'Find file(s) (as a regex), going up the parent directories'
+    set -l dir "$argv[1]"
+    set -l file "$argv[2]"
+
+    if [ -z "$dir" ]
+        return 1
+    end
+
+    while [ "$dir" ]
+        if string match -q -r "$dir/($file)\$" -- $dir/*
```
bobthecow, line 907: You'd need to regex escape the `$dir` bit, since it's perfectly reasonable to have reserved regex characters in paths. 

You might be better off putting the matching in your $dir/* expansion instead? e.g. `$dir/{.nvmrc,.node-version,package.json}`.

hansl: I just reworked the function to accept multiple values. I wanted to be clever with the glob/regex version but it's over engineering. PTAL.

functions/fish_prompt.fish
```diff
@@ -895,9 +895,38 @@ function __bobthefish_prompt_desk -S -d 'Display current desk environment'
...
 
+function __bobthefish_prompt_find_file_up -S -d 'Find file(s), going up the parent directories'
+    set -l dir "$argv[1]"
+    set -l files $argv[2..]
+
+    if [ -z "$dir" ]
+        return 1
+    end
+
+    while [ "$dir" ]
```
bobthecow, line 906: won't this loop forever because `__bobthefish_dirname /` resolves to `/`?

[the mercurial equivalent to this](https://github.com/oh-my-fish/theme-bobthefish/blob/master/functions/fish_prompt.fish#L237-L238) has the root directory special cased.

hansl: Interesting... and fixed.

image

hansl: Also note there's a refactor we can do to use this new function instead of rolling it locally.

hansl:
This PR is ready to be merged if you have no other comment. Thank you very much for the thorough review.

bobthecow:
Thank you!

PR #318, opened 2021-12-28
Dracula theme: Fix Kubernetes colors

VictorBersy:
Hello,

By default, the Kubernetes colors are displayed like so:
image

It's hard to read, and really bright, seeking for our attention. 
The purpose of displaying Kubernetes context is just to know where our `kubectl`commands will run, so it's nice to know, but not a major information that should get all our attention in my opinion.

So I tried others colors, and either we could replace the white text with a darker color:
image

Or, what I suggest in this PR, use purple as it's less attention seeking:
image

**Original version:**
Screenshot 2021-12-28 at 14 16 44

**Version with darker text:**
Screenshot 2021-12-28 at 14 16 46

**Suggested changes:**
Screenshot 2021-12-28 at 14 16 49

Tell me what you think :)

### DIFF
--- a/functions/__bobthefish_colors.fish
+++ b/functions/__bobthefish_colors.fish
@@ -516,7 +516,7 @@ function __bobthefish_colors -S -a color_scheme -d 'Define colors used by bobthe
       set -x color_vi_mode_visual           $orange $bg --bold
 
       set -x color_vagrant                  $pink $bg --bold
-      set -x color_k8s                      $green $fg --bold
+      set -x color_k8s                      $purple $bg --bold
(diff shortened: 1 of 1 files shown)


bobthecow:
You're right, that's terrible. Purple works for me. We actually use that for several other themes.

bobthecow:
I've added k8s-context to `bobthefish_display_colors` to hopefully make this less likely to happen in the future?

VictorBersy:
Awesome @bobthecow! Thanks a lot ❤️

bobthecow:
Thank you!

PR #327, opened 2022-06-10
Add Go prompt

seanhagen:
Adds the version of [Go](https://go.dev/) as reported by `go version` as a prompt segment:

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -909,9 +909,9 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
 
     if [ "$theme_display_go_actual" = "yes" ]
         if [ "$actual_go_version" = "0" ]
-            echo -ns "(NA)"
+            echo -ns "NA"
         else                    
-            echo -ns "($actual_go_version)"
+            echo -ns "$actual_go_version"
         end
     else        
         echo -ns "$gomod_version" ' '
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -852,17 +852,14 @@ end
 function __bobthefish_prompt_golang -S -d 'Display current Go information'
     [ "$theme_display_go" = 'no' ]
     and return
-    
-    
-    set -l dir $real_pwd
-    set -l found_gomod 0
-    set -l gomod_version "0"
-    set -l gomod_file
 
     # find the closest go.mod
+    set -l gomod_version "0"
+    set -l d $real_pwd
     while not [ -z "$d" ]
         if [ -e $d/go.mod ]
-            grep "^go\ " "$gomod_file" | read __ gomod_version
+            grep "^go\ " "$d/go.mod" | read __ gomod_version
+            break
         end
 
         [ "$d" = "/" ]
@@ -870,7 +867,7 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
     end
 
     # no go.mod, not in a go project, don't display the prompt
-    if test "$found_gomod" -eq "0"
+    if [ "$gomod_version" = "0" ]
         return
     end
 
@@ -1229,6 +1226,8 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
 
     # Start each line with a blank slate
     set -l __bobthefish_current_bg
+    
+    set -l real_pwd (__bobthefish_pwd)
 
     # Status flags and input mode
     __bobthefish_prompt_status $last_status
@@ -1253,7 +1252,6 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
     __bobthefish_prompt_virtualgo
     __bobthefish_prompt_node
 
-    set -l real_pwd (__bobthefish_pwd)
 
     # VCS
     set -l git_root_dir (__bobthefish_git_project_dir $real_pwd)
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -864,6 +864,8 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
 
         [ "$d" = "/" ]
         and return
+
+        set d (__bobthefish_dirname $d)
     end
 
     # no go.mod, not in a go project, don't display the prompt
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -865,7 +865,7 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
 
         if test -f "$gomod_file"
             set found_gomod 1
-            cat "$gomod_file" | grep "^go\ " | read _gomod gomod_version 
+            grep "^go\ " "$gomod_file" | read __ gomod_version
             break
         end
 
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -853,26 +853,21 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
     [ "$theme_display_go" = 'no' ]
     and return
     
-    set -l cwd (pwd)
-    set -l dir (pwd)
+    
+    set -l dir $real_pwd
     set -l found_gomod 0
     set -l gomod_version "0"
     set -l gomod_file
 
     # find the closest go.mod
-    while not test "$dir" = "/"
-        set gomod_file "$dir/go.mod"
-
-        if test -f "$gomod_file"
-            set found_gomod 1
+    while not [ -z "$d" ]
+        if [ -e $d/go.mod ]
             grep "^go\ " "$gomod_file" | read __ gomod_version
-            break
         end
 
-        cd $dir/..
-        set dir (pwd)
+        [ "$d" = "/" ]
+        and return
     end
-    cd $cwd
 
     # no go.mod, not in a go project, don't display the prompt
     if test "$found_gomod" -eq "0"
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -36,6 +36,8 @@
 #     set -g theme_display_nix no
 #     set -g theme_display_ruby no
 #     set -g theme_display_go no
+#     set -g theme_display_go_show_wrong_version yes
+#     set -g theme_display_go_actual diff
 #     set -g theme_display_user ssh
 #     set -g theme_display_hostname ssh
 #     set -g theme_display_sudo_user yes
@@ -854,14 +856,8 @@ function __bobthefish_prompt_golang -S -d 'Display current Go information'
     set -l go_version
     if type -fq asdf
         set -l asdf_golang_version (asdf current golang 2>/dev/null)
-        or set -l asdf_golang_version "na"
-        
         set -l asdf_go_sdk_version (asdf current go-sdk 2>/dev/null)
-        or set -l asdf_go_sdk_version "na"
 
-        set -l _asdf_plugin
-        set -l asdf_go_version
-        set -l asdf_provenance
         if [ "$asdf_golang_version" != "na" ]
             echo "$asdf_golang_version" | read  _asdf_plugin asdf_go_version asdf_provenance
         else if [ "$asdf_go_sdk_version" != "na" ]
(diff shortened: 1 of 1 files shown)


functions/fish_prompt.fish
```diff
@@ -803,6 +804,22 @@ function __bobthefish_rvm_info -S -d 'Current Ruby information from RVM'
     end
 end
 
+function __bobthefish_prompt_golang -S -d 'Display current Go information'
+    [ "$theme_display_go" = 'no' ]
+    and return
+    
+    set -l go_version
+    if type -fq go
+        set go_version (go version | cut --delimiter=' ' -f 3 | cut --characters='3-6' )
```
bobthecow, line 813: first, `cut` isn't particularly portable. for BSD / macOS, this would probably have to be

```fish
go version | cut -d ' ' -f 3 | cut -c '3-6'
```

but! we can probably do this with a single builtin rather than piping to `cut` (twice!) maybe this?

```fish
go version | string match -r 'go version go(\\d+\\.\\d+)' -g
```

functions/fish_prompt.fish
```diff
@@ -1172,6 +1189,7 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
     __bobthefish_prompt_nix
     __bobthefish_prompt_desk
     __bobthefish_prompt_rubies
+    __bobthefish_prompt_golang
     __bobthefish_prompt_virtualfish
     __bobthefish_prompt_virtualgo
```
bobthecow, line 1194: should this be combined with virtualgo? e.g. the `rubies` segment uses whatever is most relevant of the many ways to define a ruby version.

bobthecow (commented):
One of the principles of bobthefish is "only show things that are interesting". This is why, for example, it omits default branch names in git. In ruby, it only shows the local ruby version if it's not the system ruby version. Is there an equivalent here? Should we ever suppress the go version?

I imagine if (like me) you have system go installed, but never actually do anything with it ever, it would just be noise to have it show up in every prompt. If that's as common a scenario as I think, we should probably either (1) not show if go version == system go, or (2) make it opt-in rather than opt-out.

Thoughts?

seanhagen:
Actually, just realized there's a better way to do this based on what you said.

Instead of always showing the version based on the `go version` command, only show the version if you're in a folder that has a `go.mod` file ( or there is a `go.mod` in a folder above the one you're in ). And instead of showing the output of `go version` show the version from `go.mod`?

And you're right about it being noise, I'm already considering turning it off and I've only had it for a few days now. I think what's more useful is showing the version of Go only if you're in a Go project ( folder with `go.mod`, or a child folder inside that project ); and changing the color if the output of `go version` is lower than what's in the `go.mod`.

That way, you can see immediately if the version of Go you have is the right one when you check out a project ( or if the version of Go you have installed/configured with a version manager like [asdf](https://asdf-vm.com/) isn't the right one for the project ).

Also, maybe a flag so that it only shows if the version of Go you have isn't high enough for the project that you're in; kind of the way the Git stuff only shows if there's changes?

functions/fish_prompt.fish
seanhagen: I didn't know `cut` wasn't part of the basic set of CLI utilities on BSD or macOS systems! The `string match` version works for me so I'll switch it to that.

functions/fish_prompt.fish
seanhagen: I'd be down with that; I'll probably also see if I can steal the "only show Ruby version when in a folder where it matters" bit from the `rubies` segment too ( more thoughts on this in [my other comment](https://github.com/oh-my-fish/theme-bobthefish/pull/327#issuecomment-1155525636) )

functions/fish_prompt.fish
bobthecow: Cut does exist, but the long option names don't work on the BSD flavor (which macOS uses). 

Regardless, using a Fish builtin wherever possible saves us processing time, so we like to keep exceptions to a minimum 🙂

bobthecow:
Yep, that sounds consistent, and like it'd be a much better experience :)

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+        set go_version $gomod_version
+    end
+
+    if test "$found_gomod" -eq "0"
+        return
+    end 
+
+    set -l actual_go_version "0"
+    if ! type -fq go
+        set no_go_installed 1
```
bobthecow: what are we doing with this variable?

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+function __bobthefish_prompt_golang -S -d 'Display current Go information'
+    [ "$theme_display_go" = 'no' ]
+    and return
+    
+    set -l go_version
+    if type -fq asdf
+        set -l asdf_golang_version (asdf current golang 2>/dev/null)
+        or set -l asdf_golang_version "na"
+        
+        set -l asdf_go_sdk_version (asdf current go-sdk 2>/dev/null)
```
bobthecow: if we already got a value above, there's no need to check here, right? save ourselves the `asdf` call?

it sacrifices a tiny bit of readability, but saves a decent amount of complexity to inline all the things :) 

```fish
if type -fq asdf
  asdf current golang 2>/dev/null | read -l __ asdf_go_version __
  or asdf current go-sdk 2>/dev/null | read -l __ asdf_go_version __
  and set go_version $asdf_go_version
end
```

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+    while not test "$dir" = "/"
+        set gomod_file "$dir/go.mod"
+
+        if test -f "$gomod_file"
+            set found_gomod 1
+            cat "$gomod_file" | grep "^go\ " | read _gomod gomod_version 
+            break
+        end
+
+        cd $dir/..
```
bobthecow, line 872: we don't need to actually `cd` around the filesystem. We just want to find and read a file, right?

check out how the `hg` implementation looks for a `.hg` directory. it's almost exactly what we want to do here.

https://github.com/oh-my-fish/theme-bobthefish/blob/332f23abd7a095d5b2c024a061af7b890a4f0c20/functions/fish_prompt.fish#L230-L242

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+    set -l gomod_version "0"
+    set -l gomod_file
+
+    # find the closest go.mod
+    while not test "$dir" = "/"
+        set gomod_file "$dir/go.mod"
+
+        if test -f "$gomod_file"
+            set found_gomod 1
+            cat "$gomod_file" | grep "^go\ " | read _gomod gomod_version
```
bobthecow, line 868: we don't ever need to pipe `cat` to `grep` :) 

```fish
grep "^go\ " "$gomod_file" | read __ gomod_version
```

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+        cd $dir/..
+        set dir (pwd)
+    end
+    cd $cwd
+
+    if [ -z "$go_version" ]
+        set go_version $gomod_version
+    end
+
+    if test "$found_gomod" -eq "0"
```
bobthecow, line 878: can we use the existence of `$gomod_version` rather than setting another variable?

functions/fish_prompt.fish
```diff
@@ -846,6 +847,85 @@ function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
...
+    set -l high_enough_version 0
+    if printf "%s\n%s"  "$gomod_version" "$actual_go_version" | sort --check=silent --version-sort
+        set high_enough_version 1
+    end
+    
+    __bobthefish_start_segment $color_virtualgo
+    echo -ns $go_glyph
+    if test "$high_enough_version" -eq "0"
+        # the version of go 
+        __bobthefish_start_segment $color_rvm
```
bobthecow: doing the glyph in one color and the rest of the segment in another feels a bit awkward (and inconsistent with, e.g. the `git` segment). let's set the whole segment to the other color?

functions/fish_prompt.fish
bobthecow: oh, sorry, i didn't realize you were still changing things :)

seanhagen:
Okay, changed it so it up a bit, here's what changed.

First up, I added two variables: `theme_display_go_show_wrong_version` and `theme_display_go_actual`. More on those in a bit.

Next, check to see if we're in a Go project by looking for a `go.mod` file. If one isn't found don't display the Go version -- like how the Git prompt works.

Next, check if there's a Go executable. Not worrying about asdf or virtualgo, just if there is a binary called `go` in the PATH. If there is a Go executable, use it to grab the current version.

After that, check if the version reported by `go` meets the required version from the `go.mod`. 

What the prompt displays depends on three things:
 - the value in `theme_display_go_show_wrong_version`
 - the value in `theme_display_go_actual`
 - the version of Go from `go version` and what's in the `go.mod`

If `theme_display_go_show_wrong_version` is set to "yes" and the version reported by `go` isn't equal to or higher than the version from the `go.mod` file the version is printed out in red ( or whatever color `color_rvm` is set to ). I thought this might be useful, as it lets the user know that the version of Go they're currently using can't ( or at least _probably_ can't ) compile this project.

After that, we check the value of `theme_display_go_actual`; it can be set to "no", "yes", or "diff". If it's set to "no", the version from the `go.mod` file is displayed. If it's set to "yes", the version from `go version` is displayed. 

If it's set to "diff", it shows just the version from `go.mod` if it matches `go version`, or the `go.mod` version **followed** by the _actual_ `go version` version. For example, if the version set in `go.mod` is "1.12" and the version reported by `go version` is "1.18" it will show "1.12 (1.18)".

With those two variables the user can control:
 - if they get a warning in the prompt if the version from `go version` isn't high enough for the project
 - if they see just the version from `go.mod`, just the version from `go version`, or both ( if they're not the same )

functions/fish_prompt.fish
seanhagen: 😅

functions/fish_prompt.fish
seanhagen: I was thinking that it could be a warning that the version of Go someone has doesn't match the version in the `go.mod` file, letting them know they might not be able to compile the project

seanhagen:
@bobthecow I think it's ready for another look now

bobthecow (commented):
so the things we're trying to solve for are:

1. am i in a go project?
2. what go version _should_ i have?
3. do i have the right go version?
4. what go version _do_ i have?

i think we can cover all of those with just three options:

- `off`: never display any golang segment.
- `on`: regular colored segment with required version, or error colored segment, also with required version.
- `verbose`: regular colored segment with required version, or "error" colored segment with required version _and actual version_.

i could also see just the golang segment, but no version number, for the non-error state in `on`. i don't do enough go development to know whether this is okay, or if we need to know the version number in a non-error state.

the basic principle is only putting high signal things in the prompt. enumerate all the cases we care about, only show interesting things (e.g. when our expectations are not met) and letting the lack of a thing be one of the states.

seanhagen:
>
>  i think we can cover all of those with just three options:
> 
>     * `off`: never display any golang segment.
>     * `on`: regular colored segment with required version, or error colored segment, also with required version.
>     * `verbose`: regular colored segment with required version, or "error" colored segment with required version _and actual version_.
> 
> i could also see just the golang segment, but no version number, for the non-error state in `on`. i don't do enough go development to know whether this is okay, or if we need to know the version number in a non-error state.
> 

I went with keeping the version number always visible, even in the "ok" state because it's a piece of information that's handy to surface -- probably for the same reasons the Ruby segment does

seanhagen:
Lemme know if there is anything else that needs tweaking or polishing!

fabiomargarido:
Hey there.
It's been almost 2 years since this has seen any activity, are there still any plans to include these changes? I personally think this is helpful and have been using it (locally modifying my git copy) for quite some time.

bobthecow:
you're right. this should have been merged a long time ago 😞

fabiomargarido:
no problems, thanks for the quick response and for integrating it 🎉

PR #338, opened 2023-03-15
Hide k8s context or namespace when they are 'default'

Nekroze:
Kept the `:` prefix for the namespace to denote that it is in fact a namespace and not a context when context is default and thus down hidden.


bobthecow:
Thanks!

PR #341, opened 2023-03-28
replace AWS_SESSION_EXPIRATION for aws-vault

ismailyenigul:
aws-vault 7.X removed  AWS_SESSION_EXPIRATION env and replaced with AWS_CREDENTIAL_EXPIRATION
To make this theme works for aws-vault 7.X, we must replace this env.

see: https://github.com/99designs/aws-vault/pull/1143

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -660,13 +660,13 @@ function __bobthefish_prompt_aws_vault_profile -S -d 'Show AWS Vault profile'
     [ "$theme_display_aws_vault_profile" = 'yes' ]
     or return
 
-    [ -n "$AWS_VAULT" -a -n "$AWS_SESSION_EXPIRATION" ]
+    [ -n "$AWS_VAULT" -a -n "$AWS_CREDENTIAL_EXPIRATION" ]
     or return
 
     set -l profile $AWS_VAULT
 
     set -l now (date --utc +%s)
-    set -l expiry (date -d "$AWS_SESSION_EXPIRATION" +%s)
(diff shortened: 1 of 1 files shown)


bobthecow:
do older versions of aws-vault support `AWS_CREDENTIAL_EXPIRATION` as well, or does this change break the AWS prompt integration unless everyone has updated?

ismailyenigul:
No it is not supported in older versions. It is breaking changes in v7.0

bobthecow:
Gotcha. If that's the case it probably makes the most sense if bobthefish falls back to the old variable if the new one isn't available.

ismailyenigul:
let me try to add conditions to support AWS_CREDENTIAL_EXPIRATION as well

ismailyenigul:
what do you think about the following changes

```
[ -n "$AWS_VAULT" -a -n "$AWS_CREDENTIAL_EXPIRATION" -o "$AWS_SESSION_EXPIRATION" ]
    or return

    set -l profile $AWS_VAULT
    set -l now (gdate --utc +%s)

    if [ -n  "$AWS_CREDENTIAL_EXPIRATION" ]
        set  aws_vault_expiry (gdate -d "$AWS_CREDENTIAL_EXPIRATION" +%s)
    else
        set  aws_vault_expiry (gdate -d "$AWS_SESSION_EXPIRATION" +%s)
    end
    
    #set -l expiry (gdate -d "$AWS_CREDENTIAL_EXPIRATION" +%s)
    set -l diff_mins (math "floor(( $aws_vault_expiry - $now ) / 60)")
   
```

bobthecow:
@ismailyenigul thanks for this! i merged the fix in #351 because I missed that this one came first 😬

PR #343, opened 2023-05-08
Fix error when k8s context defined but namespace isn't

Nekroze:
Fixed simple mistake of not double qouting namespace so when empty -z consumed -o as its input.


bobthecow:
Thanks!

Issue #29, opened 2016-03-25
Git work-tree support


KishanBagaria:
Can you add an option to not show the git work-tree and instead show the working directory exactly like other themes?

This is what I see:
<img width="611" alt="screen shot 2016-03-25 at 10 44 03 pm" src="https://cloud.githubusercontent.com/assets/1093313/14049464/777eb972-f2db-11e5-95ff-22a12b1e459a.png">

The working directory isn't shortened to ~/D/K/W/k which makes it eat a lot of width.

Also, what do you think about making it look like this for all users:

`~/D/K/W/k > master > _site`

This includes all three things and makes it more consistent because the working directory is the first thing like always, then the branch, and finally the relative work-tree.

Love the theme btw, it's amazing! :+1:

bobthecow:
Hmm. It shouldn't look like that at all. 

Do you have symlinks or something else in a parent directory? Can you show me the value of $PWD and $HOME?

bobthecow:
Oh. You're using a non-standard git dir or git work tree?

KishanBagaria:
Non-standard git work tree.

```
$PWD  = /Users/kishan/Dropbox/Kishan/Websites/kishanbagaria
$HOME = /Users/kishan
```

bobthecow:
For any given PWD, it should always be `~/A/b/c/ProjectName > master > path/from/project/root`. I don't use non-standard git dir or work tree, so I haven't run into this, but your screenshot is definitely wrong. I'll take a look :)

KishanBagaria:
Repro:
```
mkdir -p git-bug/work
git-bug/
git init
echo "    worktree = ../work/" >> .git/config
trash ~/git-bug
```
<img width="822" alt="screen shot 2016-03-26 at 2 38 01 pm" src="https://cloud.githubusercontent.com/assets/1093313/14059205/68a6a8d0-f360-11e5-9ce0-58f10d7f44da.png">

What's your colour scheme btw?

bobthecow:
To be honest, I'm not sure what would be the most useful output here. If possible, it should show the `> master >` segment at the worktree root, but it's entirely possible to be inside a git dir and not share any parent paths at all with the worktree.

What would you expect to see in these cases?

KishanBagaria:
The full path in that case, but I think that's an unlikely scenario.

bobthecow:
Or even something like:

```
git --git-dir ~/git/foo --work-tree ~/Projects/foo status
[...]

PR #345, opened 2023-06-16
Fix pwd calculation for git worktrees

### LINKED ISSUE #249: `theme_git_worktree_support yes` breaks git features
Hi, I've updated omf and all plugins after _very long time_ and git features stopped working. Prompt remains like there is no git repository even though there is. Behavior remains even after disabling my `bobthefish_colors` function.
After bisecting, commit introducing this problem is 5273c92fa56cc20338ab4a834ef07427acee07b7.
```
ArchLinux
Fish version: 3.1.0
Oh My Fish version 7-1-g4ff9981
```
Anybody else had/has similar problem?

tommyknows:
A refactor changed this block from a `set git_dir $PWD` to `set git_dir $real_pwd`, but as `real_pwd` is a "static" variable and not dynamically computed, this breaks within the `pushd` / `popd` flow.

Instead of using `$real_pwd`, we need to compute the *current* directory by calling `__bobthefish_pwd` again.

Fixes #249.

---

btw, I'm not 100% certain about this worktree-support anyways. For me, it doesn't "work", or rather, I'm using worktrees but the normal git-prompt works fine.

My use-case is different from what was described in the [original issue for worktrees](https://github.com/oh-my-fish/theme-bobthefish/issues/29), and have the following directory structure:
```
$ git worktree list
/tmp/repo/main                     562496718 [main]
/tmp/repo/branch123                2fc53b215 [branch123]
/tmp/repo/another-branch           5615bf3bb [another-branch]
```

but the `.git` repo is in `tmp/repo/main/.git`. 

So my current git-prompts look like this:
```
/p/t/r/branch123 || branch123 || 
```
Meaning the branch is shown twice, once in the git-info and once in the path, while I have no idea in what project I'm in :) 

In my case, I'd rather have the left part of the prompt show me the project, with the "branch directory" trimmed away, as that's show in the git-info already. 

I don't think there's a generic solution to how people use worktrees, at least I can't see a common theme between #29 and my use-case, so I might have a look at making this functionality customisable _somehow_, although I'm not quite sure yet how to do that :) 

Thanks a lot for this great theme & prompt!

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -185,7 +185,7 @@ function __bobthefish_git_project_dir -S -a real_pwd -d 'Print the current git p
     or return
 
     pushd $git_dir
-    set git_dir $real_pwd
+    set git_dir (__bobthefish_pwd)
     popd
 
     switch $real_pwd/


bobthecow:
Good catch, and thanks for the fix!

To be honest, I don't use git worktrees at all, so I'm looking to contributors for input on how exactly it should function. If there's a way to support the original use case and yours as well, especially in a more bobthefishy way, I'm open to suggestions :)

tommyknows:
Not sure, I've got a [working version that enables my use-case](https://github.com/tommyknows/theme-bobthefish/tree/custom-worktree), but it's probably super brittle in all other use-cases 😄 

I might spend some more time on the weekend to check if it can be generalised...

PR #346, opened 2023-06-20
Make displaying python version glyph in virtualenv segment configurable

w-biggs:
Motivation: On my machine, `python --version` is taking 664ms to execute; this is then getting tacked onto every single shell command due to `__bobthefish_prompt_virtualfish` calling `python --version` to decide which version glyph to display. I'm not sure if `python --version` is simply not supposed to take that long, but I find it useful to have the current virtualenv displayed regardless, so making this configurable seems like the best route.

`$theme_display_virtualenv_python_version` is checked twice because I figured it was best to preserve the current behavior of "if the python version doesn't match any case, don't display the virtualenv color and glyph at all".


bobthecow:
So with this change it shows the glyph if you're in a virtualenv, but no version number?

For other similar things we change the binary yes/no to include an option like `verbose`. I'd be inclined to do that here, rather than introduce a new variable just to remove the version.

PR #349, opened 2023-10-04
Add AWS_VAULT to prompt - option

s256:
In addition to the AWS_VAULT profile, it's also handy to see the currently active AWS Profile. It's not mutually exclusive.
But not everyone uses `aws-vault` but e.g. granted.dev and `assume`. Which only set's the profile and does not use `aws-vault`. A dedicated setting can ensure that these two don't clash and are disabled by default

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -1212,3 +1212,6 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
 
     __bobthefish_finish_segments
 end
+if set -q VIRTUAL_ENV
+    echo -n -s (set_color -b blue white) "(" (basename "$VIRTUAL_ENV") ")" (set_color normal) " "
+end
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -656,17 +656,27 @@ end
 # Cloud Tools
 # ==============================
 
-function __bobthefish_prompt_aws_vault_profile -S -d 'Show AWS Vault profile'
-    [ "$theme_display_aws_vault_profile" = 'yes' ]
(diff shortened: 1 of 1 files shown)


bobthecow:
is there a way to make it do the right thing with whichever combination of vault and profile a user is currently using, rather than requiring manual configuration? i'm not opposed to a combined "aws" segment, similar to the way we group multiple ruby, python, or node version managers into one segment.

s256:
Haven't looked into that as I'm not using aws-vault. I could only assume the intended behaviour. Like: if the vault env var is present, than the profile should not be shown. If not, show the content of the profile env var?

bobthecow:
To be honest I'm not using either AWS option, so I'd defer what's appropriate to people who do :)

s256:
@bobthecow I merged the two functions, kept both options. But AWS Vault will take precedence, as users of `aws-vault` will get the profile info by the `$AWS_VAULT` env var. So only if that env is not set, we use `$AWS_PROFILE`.

I did stick with the new option `theme_display_aws_profile` as for me, a non-aws-vault user, I would not expect to find the setting I was looking for in the original option.

s256:
Hi bob. just wanted to follow up here :) let me know if something is missing.

bobthecow:
@s256 Sorry about the delay!

I'm thinking we collapse the two options into single `theme_display_aws_profile` (and treat `theme_display_aws_vault_profile` as a synonym for backwards compatibility). 

It looks like aws-vault has switched over to `AWS_CREDENTIAL_EXPIRATION` instead of `AWS_SESSION_EXPIRATION`, which comes from aws-cli maybe? Does the way you use aws profile also need to change? See #350 / #351

bobthecow:
(i'm happy to make whatever changes are needed, i just want to confirm what works for you!)

PR #357, opened 2024-02-09
Add option to hide python version

MaxSchambach:
This adds a simple option (via `set -g theme_hide_python_version yes`) to hide the Python version using virtualenv.

Personally, I believe most people only develop in Python 3 nowadays, so having the extra space just to show the version seems unnecessary.

PS: thanks for this awesome theme! 😀

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ This theme is based loosely on [agnoster][btf-agnoster].
 
 ### Configuration
 
-You can override some of the following default options in your `config.fish`:
+You can override some of the following default options in your `config.fish` or `.conf/omf/init.fish`:
 
 ```fish
 set -g theme_display_git no
@@ -83,6 +83,7 @@ set -g theme_display_docker_machine no
 set -g theme_display_k8s_context yes
 set -g theme_display_hg yes
 set -g theme_display_virtualenv no
+set -g theme_hide_python_version yes
 set -g theme_display_nix no
 set -g theme_display_ruby no
 set -g theme_display_node yes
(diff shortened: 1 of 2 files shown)


bobthecow:
This is a great point. I'm wondering if there's a way we can do it without config, so that by default it's the best experience for the most people? The general principle in bobthefish is only to show _interesting_ information by default, which is why it suppresses things like "current {node,ruby,python,whatever} version" if you're not using some sort of version manager or local env managed version.

Possibly we could do one of the following:

- Only show the version glyph if the current version is different from the system Python version.
- Never show the version glyph for Python 3, only Python 2.
- Add a third value (`verbose`?) to the `theme_display_virtualenv` to opt into showing the version glyph.

I'm leaning toward a combination of the last two options. By default, only show Python 2 glyph (because that's interesting, in 2024!) but if `theme_display_virtualenv` is set to `verbose` rather than `yes` we show it for Python 3 as well.

Would that take care of your concern?

MaxSchambach:
Great, I agree! I'm also more in favor of the last two points as the system's Python version is not really relevant for most developers using pyenv/virtualenv or conda to never actually touch the system one.
Showing the version of Python 2 makes sense (though not many people will likely be using it) and adding a `verbose` option to `theme_display_virtualenv` indeed seems more tidy as opposed to introducing another option in order to have it all in one place.

I'll give it a shot and update the PR 🚀

MaxSchambach:
Ready from my side, @bobthecow :)

PR #359, opened 2024-03-13
Fix golang version prompt when go directive in go.mod has patch version

fabiomargarido:
When the go directive in `go.mod` includes the patch version, i.e. `1.22.1` for instance, the prompt function treats it as `1.22` and fails to verify that the installed go version is high enough.


bobthecow:
Thank you!

PR #363, opened 2024-04-19
Add Fossil VCS support

demize:
Started using Fossil for a project, and was annoyed that I didn't have status in my prompt. This commit adds it in!

I will admit I'm not overly familiar with Fossil, so this may be less than ideal (and might be missing important statuses), but in my testing it works pretty well. I added a new option to enable Fossil support (disabled by default, because most people won't need it; Fossil is pretty quick, so it shouldn't have the same issues as Mercurial), updated the README, and tried to stick to the same coding style as the existing code.

My one complaint about my own work is the default (non-NF) Fossil glyph, which is just the bone emoji. The NF bone emoji is alright (though I'd prefer something *more* evocative of "fossil") but I couldn't think of a better symbol outside NF.

### DIFF
--- a/functions/__bobthefish_glyphs.fish
+++ b/functions/__bobthefish_glyphs.fish
@@ -50,7 +50,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
   set -x git_plus_minus_glyph '±'
 
   # Fossil glyph (it reuses most of the git glyphs)
-  set -x fossil_glyph '🦴'
+  set -x fossil_glyph \U1F9B4 # Unicode bone emoji
 
   # Disable Powerline fonts (unless we're using nerd fonts instead)
   if [ "$theme_powerline_fonts" = "no" -a "$theme_nerd_fonts" != "yes" ]
@@ -86,7 +86,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set git_stashed_glyph    \uF0C6 '' # nf-fa-paperclip
     set git_untracked_glyph  \uF128 '' # nf-fa-question
     # set git_untracked_glyph  \uF141 '' # nf-fa-ellipsis_h
-    set fossil_glyph '󰂹'
+    set fossil_glyph \UF00B9 # nf-md-bone
 
     set git_ahead_glyph      \uF47B # nf-oct-chevron_up
     set git_behind_glyph     \uF47C # nf-oct-chevron_down
@@ -99,5 +99,6 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
   # Avoid ambiguous glyphs
   if [ "$theme_avoid_ambiguous_glyphs" = "yes" ]
     set git_untracked_glyph '...'
+    set fossil_glyph '' # blank, for lack of a good fallback
   end
 end
--- a/functions/__bobthefish_glyphs.fish
+++ b/functions/__bobthefish_glyphs.fish
@@ -49,6 +49,9 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
   set -x git_minus_glyph      '-'
   set -x git_plus_minus_glyph '±'
 
+  # Fossil glyph (it reuses most of the git glyphs)
+  set -x fossil_glyph '🦴'
+
   # Disable Powerline fonts (unless we're using nerd fonts instead)
   if [ "$theme_powerline_fonts" = "no" -a "$theme_nerd_fonts" != "yes" ]
     set private_glyph           \u29B8 ' '
@@ -83,6 +86,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set git_stashed_glyph    \uF0C6 '' # nf-fa-paperclip
     set git_untracked_glyph  \uF128 '' # nf-fa-question
     # set git_untracked_glyph  \uF141 '' # nf-fa-ellipsis_h
+    set fossil_glyph '󰂹'
 
     set git_ahead_glyph      \uF47B # nf-oct-chevron_up
     set git_behind_glyph     \uF47C # nf-oct-chevron_down
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -72,6 +72,11 @@ function __bobthefish_git_branch -S -d 'Get the current git branch (or commitish
     end
 end
 
+function __bobthefish_fossil_branch -S -d 'Get the current fossil branch'
+    set -l branch (command fossil branch 2>/dev/null)
+    echo "$branch_glyph $branch"
+end
+
 function __bobthefish_hg_branch -S -d 'Get the current hg branch'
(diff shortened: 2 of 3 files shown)


functions/__bobthefish_glyphs.fish
```diff
@@ -83,6 +86,7 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
     set git_stashed_glyph    \uF0C6 '' # nf-fa-paperclip
     set git_untracked_glyph  \uF128 '' # nf-fa-question
     # set git_untracked_glyph  \uF141 '' # nf-fa-ellipsis_h
+    set fossil_glyph '󰂹'
```
bobthecow: for nerd font glyphs, please use the escaped unicode code point rather than the literal character, and add the NF name as a comment. this makes it much easier to update when things change :)

functions/__bobthefish_glyphs.fish
```diff
@@ -49,6 +49,9 @@ function __bobthefish_glyphs -S -d 'Define glyphs used by bobthefish'
   set -x git_minus_glyph      '-'
   set -x git_plus_minus_glyph '±'
 
+  # Fossil glyph (it reuses most of the git glyphs)
+  set -x fossil_glyph '🦴'
```
bobthecow: this should probably be suppressed in the `theme_avoid_ambiguous_glyphs` block below. that's used for terminals that don't love modern characters and can't deal with things like emojis. i'm not sure there's a good non-emoji fallback, so maybe just have it be empty string?

functions/fish_prompt.fish
```diff
@@ -1020,6 +1037,54 @@ end
 # VCS segments
 # ==============================
 
+function __bobthefish_prompt_fossil -S -a fossil_root_dir -a real_pwd -d 'Display the actual fossil state'
+    set fossil_statuses (command fossil changes --differ 2>/dev/null | cut -d' ' -f1 | sort -u)
```
bobthecow: prolly should use `set -l` or `-f` for all of these, to avoid contaminating user variables from outside the function scope.

functions/fish_prompt.fish
```diff
@@ -1020,6 +1037,54 @@ end
...
+    if [ "$dirty" ]
+        set flag_colors $color_repo_dirty
+    end
+
+    __bobthefish_path_segment $fossil_root_dir project
+
+    __bobthefish_start_segment $flag_colors
+    echo -ns $fossil_glyph ' '
+
+    __bobthefish_start_segment $flag_colors
```
bobthecow: do you want to start a new segment or just prefix the current segment with a bone? it's probably more bobthefishy to keep it in a single segment?

functions/fish_prompt.fish
```diff
@@ -1255,6 +1321,8 @@ function fish_prompt -d 'bobthefish, a fish theme optimized for awesome'
         __bobthefish_prompt_git $git_root_dir $real_pwd
     else if [ "$hg_root_dir" ]
         __bobthefish_prompt_hg $hg_root_dir $real_pwd
+    else if [ "$fossil_root_dir" ]
+        __bobthefish_prompt_fossil $fossil_root_dir $real_pwd
```
bobthecow: i dunno if there's a better way to do this, but for git+hg i added a naive nested repo support. doing this three ways feels like it'd be a mess.

maybe we could do something with `string length` of the various paths. or maybe it's not even worth worrying about.

demize:
```fish
switch (path sort -r "$git_root_dir" "$hg_root_dir" "$fossil_root_dir")[1]
  case "$git_root_dir"
    __bobthefish_prompt_git $git_root_dir $real_pwd
  case "$hg_root_dir"
    __bobthefish_prompt_hg $hg_root_dir $real_pwd
  case "$fossil_root_dir"
    __bobthefish_prompt_fossil $fossil_root_dir $real_pwd
end
```

I think this would work, and is probably least messy. Not sure if it's worth it either, but I think this is an elegant enough solution. And it could replace the entire series of ifs/elses, since `path sort` will handle empty paths as if they're above `/`:

```shell
$ path sort -r "git" "" "git/fossil repo"
git/fossil repo
git

$ # previous line intentionally left blank
```

bobthecow: Yeah I don't hate this approach. It's definitely easier to follow than the nested ifs and switches.

Threw it on the old `--profile` and it looks like `path sort` for two VCS paths is about a 3x penalty vs ifs and cases and string matching, and a 2x penalty when comparing three paths. I think that's something we can afford.

Let's do this.

bobthecow: Note that the "no VCS directory at all" case has to be handled explicitly, otherwise `path sort` will return an empty string, and it'll match the first `case "$git_root_dir"` (which itself is an empty string too). So something like this:

```fish
switch (path sort -r "$git_root_dir" "$hg_root_dir" "$fossil_root_dir")[1]
    case ''
        __bobthefish_prompt_dir $real_pwd
    case "$git_root_dir"
        __bobthefish_prompt_git $git_root_dir $real_pwd
    case "$hg_root_dir"
        __bobthefish_prompt_hg $hg_root_dir $real_pwd
    case "$fossil_root_dir"
        __bobthefish_prompt_fossil $fossil_root_dir $real_pwd
end
```

functions/__bobthefish_glyphs.fish
demize: Used the escaped code point here, and for the bone emoji too :)

functions/__bobthefish_glyphs.fish
demize: Tried both `''` and `' '`, went with the null string because it looks better.

functions/fish_prompt.fish
demize: Good point; much better as a single segment.

demize:
Addressed all the feedback, tweaked the prompt a little bit more, and also noticed that sometimes Fossil would take 1-2 seconds to print the `fossil info` output. Tracking that down and coming up with a solution took me longer than anticipated, but I got something pretty good working with `fossil json status` I think!

functions/fish_prompt.fish
```diff
@@ -197,14 +197,15 @@ end
...
-    and command fossil status >/dev/null 2>/dev/null
+    and command fossil json status >/dev/null 2>/dev/null
     or return
 
     set -q theme_vcs_ignore_paths
     and [ (__bobthefish_ignore_vcs_dir $real_pwd) ]
     and return
 
-    command fossil info 2>/dev/null | sed '3q;d' | string split ' ' -f2 -n | string trim --right --chars=/
+    set -f dir (command fossil json status 2>/dev/null | grep localRoot | string split ':' -f2 | string trim --chars='"/,')
```
bobthecow: rather than calling it a second time, should we save the output from the start of the function and use that?

demize: Yeah, that makes more sense than what I was doing... just inlined the `set` into the first check, which works nicely.

bobthecow:
Awesome, thanks for this!

PR #367, opened 2024-06-18
Update README with better custom color scheme instructions

frothedoatmilk:
Provide an up to date description of how to override colors for a theme. Ripped the code verbatim from the error you get when trying to use `set -g theme_color_scheme user`, which is in turn from functions/__bobthefish_colors.fish


bobthecow:
Thanks!

PR #373, opened 2024-07-23
remove unknown '-g' flag from 'string match' command

### LINKED ISSUE #372: Go version display broken
When checking out in a go project directory, I have the following error:

```bash
string match: Unknown option “-g”

~/.local/share/omf/themes/bobthefish/functions/fish_prompt.fish (line 1): 
go version | string match -r 'go version go(\\d+\\.\\d+(?:\\.\\d+)?)' -g
^
in command substitution
	called on line 876 of file ~/.local/share/omf/themes/bobthefish/functions/fish_prompt.fish
in function '__bobthefish_prompt_golang' with arguments '/home/air1bzz/code/binary_clock'
	called on line 1314 of file ~/.local/share/omf/themes/bobthefish/functions/fish_prompt.fish
in function 'fish_prompt'
in command substitution

(Type 'help string' for related documentation)
```

It is because of the call of function `__bobthefish_prompt_golang`,
Display is broken because of the unknown `-g` flag passed to the `string match` command.

air1bzz:
Fixes #372

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -873,7 +873,7 @@ function __bobthefish_prompt_golang -S -a real_pwd -d 'Display current Go inform
     set -l actual_go_version "0"
     set -l high_enough_version "0"
     if command -q go
-        set actual_go_version (go version | string match -r 'go version go(\\d+\\.\\d+(?:\\.\\d+)?)' -g)
+        set actual_go_version (go version | string match -r 'go version go(\\d+\\.\\d+(?:\\.\\d+)?)')
         if printf "%s\n%s"  "$gomod_version" "$actual_go_version" | sort --check=silent --version-sort
             set high_enough_version "1"
         end


bobthecow:
[`-g` has been part of fish since 3.4.0](https://github.com/fish-shell/fish-shell/releases/tag/3.4.0) so it's not exactly "unknown" 😛

in this case, removing it isn't the same thing. `-g` with `string match` returns only the regex capturing groups. so the code in question selects just the version number and stores it as a variable.

`go_version` will be something like...

```
go version go1.23.0 darwin/arm64
```

then `go version | string match -r 'go version go(\\d+\\.\\d+(?:\\.\\d+)?)' -g` return the version:

```
1.23.0
```

… but leaving off the `-g` returns the whole match, first, and the capturing groups second:

```
go version go1.23.0
1.23.0
```

… which is then passed to the `sort` on the next line, and it silently compares a go version with the string `go version ...`, which isn't going to work at all 🙂

bobthecow:
I've pushed a change that uses `string replace` rather than `string match`, and confirmed the required options are available on `string` all the way back to fish 2.0.0. Mind confirming that it works for you?

air1bzz:
Hello, thanks you for explanations and having treated this case.
Sincerely,

bobthecow:
No problem. Thanks for reporting!

PR #378, opened 2026-02-03
Enhance support for asdf version management in README and prompt functions

benklop:
- Updated README.md to include asdf support for Ruby, Node.js, and Python versions.
- Modified fish_prompt.fish to display asdf-managed Python and Node.js versions when no virtualenv or conda environment is active.
- Adjusted bobthefish_display_colors.fish to include asdf in the color display for Node.js.


benklop:
Still working on / testing this

PR #379, opened 2026-02-12
Add support for mise and refactor asdf version detection for Ruby and Node

Rykian:
This PR introduces support for the [mise](https://mise.jdx.dev/) version manager and refactors the existing asdf logic into reusable helper functions to reduce code duplication.

Changes:

- New Helper Functions:
  - `__bobthefish_asdf_plugin_version`: Extracted the asdf parsing logic that was previously inline. It ensures we continue to ignore versions set in `$HOME/.tool-versions` (global).
  - `__bobthefish_mise_plugin_version`: specific logic for mise. It parses mise `ls --current` to check the source of the version and ignores it if it comes from the global config (`~/.config/mise/config.toml`), matching the theme's existing behavior for other managers.
- Ruby Prompt:
  - Updated to use the new asdf helper.
  - Added detection for mise.
- Node Prompt:
  - Added asdf and mise to the manager detection list.
  - Refactored the version retrieval logic to handle the specific requirements of asdf/mise (via helpers) vs nvm/fnm (via current command).

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -813,6 +813,22 @@ function __bobthefish_asdf_plugin_version -S -d 'Guess version of an executable 
     echo $asdf_plugin_version
 end
 
+function __bobthefish_mise_plugin_version -S -d 'Guess version of an executable managed by mise'
+    set -l mise_version (mise current $argv[1] 2>/dev/null)
+    or return
+
+    # mise outputs just the version number, unlike asdf which outputs "tool version source"
+    # Only show if not from global config
+    set -l mise_source (mise ls --current $argv[1] 2>/dev/null | tail -n 1 | awk '{print $3}')
+    
+    # Don't show if it's from the global config (expand ~ to $HOME for comparison)
+    set -l global_config (string replace '~' "$HOME" "$mise_source")
+    [ "$global_config" = "$HOME/.config/mise/config.toml" ]
+    and return
+
+    echo $mise_version
+end
+
 function __bobthefish_prompt_rubies -S -d 'Display current Ruby information'
     [ "$theme_display_ruby" = 'no' ]
     and return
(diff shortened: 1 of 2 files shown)


functions/fish_prompt.fish
```diff
@@ -800,6 +800,35 @@ function __bobthefish_rvm_info -S -d 'Current Ruby information from RVM'
     end
 end
 
+function __bobthefish_asdf_plugin_version -S -d 'Guess version of an executable managed by asdf'
+    set -l asdf_current_plugin (asdf current $argv[1] 2>/dev/null)
```
bobthecow: let's follow the repo pattern of preferring named function args?

functions/fish_prompt.fish
```diff
@@ -1005,20 +1027,40 @@ function __bobthefish_prompt_node -S -d 'Display current node version'
...
+      set node_manager_dir $NVM_DIR
+    else if type -fq fnm
+      set node_manager 'fnm'
+      set node_manager_dir $FNM_DIR
+    else if type -fq asdf
+      set node_manager 'asdf'
+      set node_manager_dir $ASDF_DIR
+    else if type -fq mise
+      set node_manager 'mise'
+      set node_manager_dir ~/.config/mise/
```
bobthecow: i worry a little bit about using the presence of a tool to decide that _this_ is the tool users use to manage their $whatever version. especially for things like asdf and mise where they handle multiple things. maybe you use asdf for node, but chruby for ruby or something.

is there something smarter we could do here? maybe combine the `type -q` checks with parsing the version for that specific runtime, and continuing to try the next manager if you don't get a version for the previous.

or maybe we use naive heuristics and let users configure overrides.

bobthecow (changes requested):
thank you! i left a couple of comments inline.

Rykian:
Sorry for the delay, I needed to recalibrate my brain to write fish code, don't use it often.
If you need modifications, tell me!

functions/fish_prompt.fish
Rykian: You're right, it's done!

functions/fish_prompt.fish
Rykian: I used multiple condition in my last commit to check every version manager until a version is not null and added a theme configuration variable. Hope my code is ok.

